PR #21, opened 2021-09-27
Muairss write

Jo-stfc:
muairss write changes


pm_muairss_write/get_out_folder.py
```diff
@@ -0,0 +1,27 @@
...
+                    (exception == '' or exception not in i[0])):
+                filelst.append(i[0]+os.sep+j)
+    return filelst
+
+
+try:
+    with open('params.yaml') as f:
+        data = yaml.safe_load(f)
+except FileNotFoundError:
+    sys.stdout.write('default')
```
elichad, line 23: 1. Why the choice of `sys.stdout.write` over `print` here? I'm unfamiliar with the former - does it just avoid a trailing `\n` or are there other reasons to use it?
2. If `out_folder` is not provided, `pm-muairss -t w` [defaults to](https://github.com/muon-spectroscopy-computational-project/pymuon-suite/blob/master/pymuonsuite/schemas.py#L173) the folder name `muon-airss-out`, so this line should reflect that. Currently it causes an error as no folder `default` exists to put into the zip.
3. We should add a test that includes a parameter file that does not include `out_folder` to catch this problem in future.

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -3,27 +3,43 @@
...
+            <option value="yes">yes</option>
+        </param>   
+        <when value="yes">
+            <param type="data" name="castep_param" format="txt" />
+        </when>
+    </conditional>
     </inputs>
     <outputs>
-        <data label="$out_folder" name="out_zip" format="zip" from_work_dir="out_zip.zip"/>
+        <data label="muonated" name="out_zip" format="zip" from_work_dir="out_zip.zip"/>
```
elichad, line 35: It would be nice to have a more descriptive label, like "Muonated structures for $structure" or something else that varies with the structure you put in.

pm_muairss_write/get_out_folder.py
```diff
@@ -0,0 +1,27 @@
+import os
```
elichad, line 1: You can't add comments to empty files so I'm putting it here: does the `initial` file above do anything?

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -3,27 +3,43 @@
...
+        ln -s out_zip1.zip out_zip.zip &&
+        find \$out_folder > tree.txt
     ]]></command>
     <inputs>
         <param type="data" name="structure" format="cell" />
-        <param type="data" name="params" format="yaml" />
-        <param type="text" name="out_folder"/>
+        <param type="data" name="params" format="txt" />
+    <conditional name="iscastep">
+        <param type="select" value="no" name="iscastep_sel" label="Is the calculator castep?">
```
elichad, line 25: This parameter looks like it would work better as type [`boolean`](https://docs.galaxyproject.org/en/master/dev/schema.html#boolean) rather than `select`.

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -3,27 +3,43 @@
...
+        ln -s $params params.yaml &&
+        #if str($iscastep.iscastep_sel)=="yes":
+            ln -s $castep_param castep.param &&
+        #end if
         pm-muairss -t w input.cell params.yaml &&
-        zip -r out_zip.zip $out_folder &&
-        find $out_folder > tree.txt
+        out_folder="`python ${__tool_directory__}/get_out_folder.py`" &&
+        #if str($iscastep.iscastep_sel)=="yes":
+            zip -r out_zip1.zip \$out_folder params.yaml input.cell castep.param &&
```
elichad, line 14: It'd be nice to use the original input file names here if we can to avoid confusion if people unzip the files, but I'm undecided if this is truly necessary. What do you think?

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -3,27 +3,43 @@
...
+        #else if str($iscastep.iscastep_sel)=="no":
+            zip -r out_zip1.zip \$out_folder params.yaml input.cell &&
+        #end if
+        ln -s out_zip1.zip out_zip.zip &&
+        find \$out_folder > tree.txt
     ]]></command>
     <inputs>
         <param type="data" name="structure" format="cell" />
```
elichad, line 22: These parameters could all do with help text etc., but this applies to all tools and I think we can make it a separate issue.

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -3,27 +3,43 @@
...
-        cp $params params.yaml &&
+        ln -s $structure input.cell &&
+        ln -s $params params.yaml &&
+        #if str($iscastep.iscastep_sel)=="yes":
+            ln -s $castep_param castep.param &&
+        #end if
         pm-muairss -t w input.cell params.yaml &&
-        zip -r out_zip.zip $out_folder &&
-        find $out_folder > tree.txt
+        out_folder="`python ${__tool_directory__}/get_out_folder.py`" &&
```
elichad, line 12: Nice work figuring out all the bizarre syntax to get this command block working!

elichad (changes requested):
This is a great start. Most comments are minor for tidiness, with just a couple of bigger suggestions.

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -3,27 +3,43 @@
...
-        <data label="$out_folder" name="out_zip" format="zip" from_work_dir="out_zip.zip"/>
+        <data label="muonated" name="out_zip" format="zip" from_work_dir="out_zip.zip"/>
         <data label="file tree for testing" name="file_tree" format="txt" from_work_dir="tree.txt" hidden="true"/>
     </outputs>
     <tests>
         <test>
             <param name="structure" value="Si.cell" ftype="txt"/>
             <param name="params" value="Si-muairss-castep.yaml" ftype="txt"/>
-            <param name="out_folder" value="si-muon-airss-out"/>
-            <output name="out_zip" file="si-muon-airss-out.zip" ftype="zip" compare="sim_size" delta_frac="0.1"/>
```
elichad, line 26: Thought of one more thing - we should remove this zip folder referenced here from `test-data`, plus anything else that's no longer used for testing.

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -3,27 +3,43 @@
...
         pm-muairss -t w input.cell params.yaml &&
-        zip -r out_zip.zip $out_folder &&
-        find $out_folder > tree.txt
+        out_folder="`python ${__tool_directory__}/get_out_folder.py`" &&
+        #if str($iscastep.iscastep_sel)=="yes":
+            zip -r out_zip1.zip \$out_folder params.yaml input.cell castep.param &&
+        #else if str($iscastep.iscastep_sel)=="no":
+            zip -r out_zip1.zip \$out_folder params.yaml input.cell &&
+        #end if
+        ln -s out_zip1.zip out_zip.zip &&
```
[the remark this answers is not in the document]
Jo-stfc: I was trying to set up the output as a symlink, but there doesnt seem to be much difference in execution time, so i might as well keep the output as the original file

pm_muairss_write/get_out_folder.py
Jo-stfc: no, it was just to resolve some conflict occuring in creating a new branch(needed initial commit changes)

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -3,27 +3,47 @@
...
         <test>
             <param name="structure" value="Si.cell" ftype="txt"/>
             <param name="params" value="Si-muairss-castep.yaml" ftype="txt"/>
-            <param name="out_folder" value="si-muon-airss-out"/>
-            <output name="out_zip" file="si-muon-airss-out.zip" ftype="zip" compare="sim_size" delta_frac="0.1"/>
+            <output name="file_tree" file="tree.txt" ftype="txt" compare="sim_size" delta="350"/>
+        </test>
+        <test>
+            <param name="structure" value="Si.cell" ftype="txt"/>
+            <param name="params" value="Si-muairss-castep-2.yaml" ftype="txt"/>
```
elichad, line 46: Coming back to this after a while, I couldn't remember what this second yaml file was testing. It would be better if this file was renamed to indicate its testing purpose – I would suggest `Si-castep-without-out-folder.yaml` or something similar. We could also rename the first test file to indicate that it does have the out folder.

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -3,27 +3,47 @@
...
+            <option value="yes">yes</option>
+        </param>   
+        <when value="yes">
+            <param type="data" name="castep_param" format="txt" />
+        </when>
+    </conditional>
     </inputs>
     <outputs>
-        <data label="$out_folder" name="out_zip" format="zip" from_work_dir="out_zip.zip"/>
+        <data label="muonated $structure.name-$params.name" name="out_zip" format="zip" from_work_dir="out_zip.zip"/>
```
elichad, line 35: This is much better than before, but it will generate labels like `muonated Si.cell-Si-muairss-castep.yaml` which is a bit confusing to read with all the hyphenation. I suggest:
```suggestion
        <data label="muonated $structure.name using $params.name" name="out_zip" format="zip" from_work_dir="out_zip.zip"/>
```

elichad (changes requested):
A couple of extra comments on the changes you've made so far (both minor). 

There are still some unaddressed comments from the previous review, please make sure those are handled too.

pm_muairss_write/get_out_folder.py
```diff
@@ -0,0 +1,11 @@
+import sys
+
+import yaml
+
+
+try:
+    with open('params.yaml') as f:
+        data = yaml.safe_load(f)
+    sys.stdout.write(data['out_folder'])
+except KeyError or FileNotFoundError:
```
elichad, line 10: Oh, we should remove the FileNotFoundError from here too, I don't think it got copied from the other PRs. (Hopefully it shouldn't ever be hit since the input comes direct from the Galaxy interface in this case)

pm_muairss_write/pm_muairss_write.xml
Jo-stfc: changed

pm_muairss_write/pm_muairss_write.xml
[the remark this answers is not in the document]
elichad: I don't think there is any point doing the symlink here – for inputs it is beneficial because the files already exist in the Galaxy history and we don't have to use up extra space by copying them, but for outputs we have created something new so I don't think the symlink matters.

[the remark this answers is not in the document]
elichad: Oops, you had already done this and I was looking at the wrong commit. My bad!

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -3,27 +3,46 @@
...
     <command detect_errors="exit_code"><![CDATA[
         pip show pymuon-suite || ( git clone https://github.com/muon-spectroscopy-computational-project/pymuon-suite /tmp/pymuon-suite && pip install /tmp/pymuon-suite && rm -rf /tmp/pymuon-suite) &&
-        cp $structure input.cell &&
-        cp $params params.yaml &&
+        ln -s $structure input.cell &&
+        ln -s $params params.yaml &&
+        #if str($iscastep.iscastep_sel)=="true":
+            ln -s $castep_param castep.param &&
+        #end if
         pm-muairss -t w input.cell params.yaml &&
```
elichad, line 11: If `params.yaml` does not have `castep_param` set, pm-muairss does not pick up the `castep.param` file at runtime (in fact, this is probably true if `castep_param` is set to anything other than `castep.param`). It therefore defaults to a generic CASTEP param file and sticks that in all the output folders instead. We need to make sure this parameter file actually gets passed through to pm-muairss (by adding it to params.yaml if it's not there, or tweaking the line if it is).

This also means that no error is thrown if calculator is castep but iscastep_sel is false – perhaps this shouldn't throw an error if it does fall back on defaults, but it should maybe raise a warning.

elichad (changes requested):
I'm happy with all the code changes now, but I found a bug while testing manually.

pm_muairss_write/pm_muairss_write.xml
Jo-stfc: fixed, now muairss write overwrites the castep param config value with the loaded file. Also updated castep defaults from pymuonsuite schema

elichad: It should also add the castep_param line if it isn't already present, IMO.

What do you mean by the CASTEP defaults?

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -1,29 +1,52 @@
...
     <requirements>
+        <requirement type="package">pymuon-suite</requirement>
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
         pip show pymuon-suite || ( git clone https://github.com/muon-spectroscopy-computational-project/pymuon-suite /tmp/pymuon-suite && pip install /tmp/pymuon-suite && rm -rf /tmp/pymuon-suite) &&
-        cp $structure input.cell &&
-        cp $params params.yaml &&
+        ln -s $structure input.cell &&
+        ln -s $params params.yaml &&
+        #if str($iscastep.iscastep_sel)=="true":
```
elichad, line 9: Can these if statements be combined?

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -1,29 +1,52 @@
...
             <param name="structure" value="Si.cell" ftype="txt"/>
-            <param name="params" value="Si-muairss-castep.yaml" ftype="txt"/>
-            <param name="out_folder" value="si-muon-airss-out"/>
-            <output name="out_zip" file="si-muon-airss-out.zip" ftype="zip" compare="sim_size" delta_frac="0.1"/>
+            <param name="params" value="Si-muairss-castep-with-out-folder.yaml" ftype="txt"/>
+            <output name="file_tree" file="tree.txt" ftype="txt" compare="sim_size" delta="350"/>
+        </test>
+        <test>
+            <param name="structure" value="Si.cell" ftype="txt"/>
+            <param name="params" value="Si-muairss-castep-no-out-folder.yaml" ftype="txt"/>
```
elichad, line 49: One thing that we miss in these tests is whether the castep_param variable is correctly updated/whether the generated param files have the expected contents. Can you think of any way to test this?

elichad (changes requested):
nearly there, I promise

pm_muairss_write/pm_muairss_write.xml
Jo-stfc: https://github.com/muon-spectroscopy-computational-project/pymuon-suite/blob/master/pymuonsuite/schemas.py

pm_muairss_write/pm_muairss_write.xml
elichad: Discussed: this is a difficult test to make, and would require writing more content to test files as well even if we figured out how to approach it. We will ask the Galaxy developers if there is a way to run a script as part of testing. If there isn't, we will rely on manual testing for this aspect.

pm_muairss_write/pm_muairss_write.xml
Jo-stfc: statements combined

pm_muairss_write/pm_muairss_write.xml
Jo-stfc: edited sed to add castep param if missing.

elichad (changes requested):
Final thing, we can get this finished on Monday :)

pm_muairss_write/pm_muairss_write.xml
elichad: Resolved: use hidden parameters to write extra stuff to file

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -41,13 +46,27 @@
     <tests>
         <test>
             <param name="structure" value="Si.cell" ftype="txt"/>
-            <param name="params" value="Si-muairss-castep-with-out-folder.yaml" ftype="txt"/>
-            <output name="file_tree" file="tree.txt" ftype="txt" compare="sim_size" delta="350"/>
+            <param name="testing" value="true"/>
+            <param name="params" value="configcastep_t_with_out_folder_and_castep_param.yaml" ftype="txt"/>
```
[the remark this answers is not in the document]
Jo-stfc: added

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -41,13 +46,27 @@
...
+            <output name="file_tree" file="tree_castep_t_with_out_folder_and_castep_param.txt" ftype="txt" compare="sim_size" delta="350"/>
```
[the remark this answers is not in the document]
Jo-stfc: added

Issue #34, opened 2022-01-19
YAML tool fails with default settings


elichad:
```
{"err_msg": "parameter 'geom_steps': the attribute 'value' must be set for non optional parameters, 
             parameter 'geom_force_tol': the attribute 'value' must be set for non optional parameters, 
             parameter 'max_scc_steps': the attribute 'value' must be set for non optional parameters", 
"err_code": 400008, 
"err_data": {"general_params|geom_steps": "parameter 'geom_steps': the attribute 'value' must be set for non optional parameters", "general_params|geom_force_tol": "parameter 'geom_force_tol': the attribute 'value' must be set for non optional parameters", "clustering_params|max_scc_steps": "parameter 'max_scc_steps': the attribute 'value' must be set for non optional parameters"},
"param_errors": {"general_params|geom_steps": {"message": "parameter 'geom_steps': the attribute 'value' must be set for non optional parameters", "message_suffix": "the attribute 'value' must be set for non optional parameters", "parameter_name": "geom_steps", "parameter_value": ""}, "general_params|geom_force_tol": {"message": "parameter 'geom_force_tol': the attribute 'value' must be set for non optional parameters", "message_suffix": "the attribute 'value' must be set for non optional parameters", "parameter_name": "geom_force_tol", "parameter_value": ""}, "clustering_params|max_scc_steps": {"message": "parameter 'max_scc_steps': the attribute 'value' must be set for non optional parameters", "message_suffix": "the attribute 'value' must be set for non optional parameters", "parameter_name": "max_scc_steps", "parameter_value": ""}}
```

- dig in the pymuonsuite code to figure out if there are default values that get used that aren't set in the schema, or if they really are optional in which case add the optional tag in the XML
- make tests that run with default settings, including the defaults for the different conditionals, so this doesn't trip us up again

PR #35, opened 2022-01-20
update yamltool default values

Jo-stfc:


elichad:
Can you add the tests as mentioned in #34 please, to make sure this isn't an issue for any of the other conditional defaults

elichad (approved):
Looks good to me. I was expecting more tests for the different calculator defaults etc., but I think that as all the conditional parameters are optional there shouldn't be any problems, so we can leave it.

PR #41, opened 2022-01-28
Symmetry

Jo-stfc:


pm_symmetry/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Computational chemistry
+- Molecular Dynamics
+description: casculate structure symmetries (pm-symmetry)
```
elichad, line 4:
```suggestion
description: Calculate structure symmetries (pm-symmetry)
```

pm_symmetry/pm_symmetry.xml
```diff
@@ -0,0 +1,40 @@
+<tool id="pm_symmetry" name="Generate Wyckoff points symmetry report (pm-symmetry)" version="0.1.0" python_template_version="3.5">
+    <requirements>
+        <requirement type="package">pymuonsuite</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        pip3 show pymuonsuite || ( git clone https://github.com/muon-spectroscopy-computational-project/pymuon-suite /tmp/pymuon-suite && pip install /tmp/pymuon-suite && rm -rf /tmp/pymuon-suite) &&
```
elichad, line 6: Line no longer needed
```suggestion

```

pm_symmetry/pm_symmetry.xml
```diff
@@ -0,0 +1,40 @@
...
+        ln -s $structure input.cell &&
+        pm-symmetry input.cell > out.txt 
+    ]]></command>
+    <inputs>
```
elichad, line 10: Same comments as usual for inputs/outputs: clear names, labels and help text please. 

Remember that the `name` can appear in the workflow UI so it's not totally hidden from the user - 'structure' is a fine name, but 'out_sym' isn't that clear.

pm_symmetry/pm_symmetry.xml
```diff
@@ -0,0 +1,40 @@
...
+        <param type="data" name="structure" format="cell" />
+        <param type="hidden" name="testing" value="false"/>
```
elichad, line 12: This parameter doesn't seem to be used anywhere, so should be removed.

pm_symmetry/pm_symmetry.xml
```diff
@@ -0,0 +1,40 @@
...
+        </test>
+    </tests>
+    <help><![CDATA[
+        usage: pm-symmetry structure
+
+        Given an input structure, generates a Wyckoff points symmetry report for that structure.
+    ]]></help>
+    <citations>
+        <citation type="bibtex">
+@misc{github,
```
elichad, line 31: For this citation, same comments as in #36 .

pm_symmetry/pm_symmetry.xml
```diff
@@ -0,0 +1,40 @@
...
+            <output name="out_sym" file="test_out.txt" ftype="txt"  compare="sim_size" delta="500">
+            </output>
+        </test>
+    </tests>
+    <help><![CDATA[
+        usage: pm-symmetry structure
+
+        Given an input structure, generates a Wyckoff points symmetry report for that structure.
+    ]]></help>
+    <citations>
```
elichad, line 29: Are there any paper citations needed for `pm-symmetry`? IIRC it is described in the UEP paper

pm_symmetry/test-data/test_out.txt
```diff
@@ -0,0 +1,52 @@
+Wyckoff points symmetry report for input.cell
```
elichad, line 1: This one is pedantic, but it would be nice if you could get `input.cell` in these outputs to be replaced by the input filename - i.e. `Si.cell` in this case. Prevents confusion.

elichad: Actually, thinking about it, perhaps we don't even need to use the `input.cell` placeholder in the tool command - just symlink to `$structure.name`? Perhaps you can skip the symlink altogether, but I don't know if that would result in some long absolute path being output, which we'd want to avoid.

elichad (changes requested):
Glad we have a small and uncomplicated tool for once!

pm_symmetry/test-data/test_out.txt
Jo-stfc: updated to use $structure.name

pm_symmetry/pm_symmetry.xml
elichad: Structure parameter needs a label and some help text still - can be taken from pm_muairss_write

pm_symmetry/pm_symmetry.xml
```diff
@@ -0,0 +1,57 @@
+<tool id="pm_symmetry" name="Generate Wyckoff points symmetry report (pm-symmetry)" version="0.1.0" python_template_version="3.5">
+    <requirements>
+        <requirement type="package">pymuonsuite</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        ln -s $structure $structure.name &&
+        pm-symmetry $structure.name > out.txt 
+    ]]></command>
+    <inputs>
+        <param type="data" name="structure" label="Structure file" format="cell" help="The structure to add muons to. The original file will be preserved. Accepted file types: cell."/>
```
elichad, line 10:
```suggestion
        <param type="data" name="structure" label="Structure file" format="cell" help="The structure to calculate symmetries of. Accepted file types: cell."/>
```

elichad (approved):

PR #47, opened 2022-02-11
Uepplot

Jo-stfc:
pm-uep-plot command tool. output is a zip of dats and the inputs


pm_uep_plot/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Computational chemistry
+- Molecular Dynamics
+description: Generate uep plots (pm-uep-plot)
+homepage_url: https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools
+long_description: |
+  Given input dft parameters (den_fmt and castep) and plotting coordinates, generate plots.
+name: pm_uep_plot
+owner: muon-spectroscopy-computational-project
+remote_repository_url: https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/main/pm_muairss_write
```
elichad, line 10: URL needs updating

pm_uep_plot/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Computational chemistry
+- Molecular Dynamics
+description: Generate uep plots (pm-uep-plot)
+homepage_url: https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools
+long_description: |
+  Given input dft parameters (den_fmt and castep) and plotting coordinates, generate plots.
```
elichad, line 7:
```suggestion
  Given outputs from a CASTEP DFT calculation (.den_fmt and .castep) and plotting coordinates, generate plots of the Unperturbed Electrostatic Potential (UEP).
```

pm_uep_plot/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Computational chemistry
+- Molecular Dynamics
+description: Generate uep plots (pm-uep-plot)
```
elichad, line 4:
```suggestion
description: Generate UEP plots (pm-uep-plot)
```

pm_uep_plot/pm_uep_plot.xml
```diff
@@ -0,0 +1,82 @@
+<tool id="pm_uep_plot" name="generate plots from charge density" version="0.1.1" python_template_version="3.5">
```
elichad, line 1: This name should match the shed file

pm_uep_plot/pm_uep_plot.xml
```diff
@@ -0,0 +1,82 @@
...
+            <param name="charge_density" value="Si.den_fmt" ftype="txt"/>
+            <param name="castep_log" value="Si.castep" ftype="txt" />
+            <param name="line_plots" value="[[1,1,1],[7,6,1],[1,2,3]]" />
+            <output name="file_tree" file="tree.txt" ftype="txt" compare="sim_size" delta="10"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+        usage: pm-uep-plot configuration
+
+        Given an input charge density, castep log and coordinate set, generates the resulting plot.
```
elichad, line 45: This line should match the shed file (IMO)

pm_uep_plot/pm_uep_plot.xml
```diff
@@ -0,0 +1,82 @@
...
+
+        |
+        | Line plot accepted formats
```
elichad, line 48: These tables are useful, but could do with some examples to go with them. Or perhaps we could construct these inputs within the tool code from smaller chunks - individual fields for crystallographic direction, starting point, etc. Smaller chunks would probably be better from a UI perspective. What do you think? We can discuss 'sensible' example inputs if needed.

pm_uep_plot/pm_uep_plot.xml
```diff
@@ -0,0 +1,82 @@
...
+        <citation type="bibtex">
+@misc{github,
+  author = {Muon Spectroscopy Computational Project},
+  year = {2021},
+  title = {pymuon-suite},
+  publisher = {GitHub},
+  journal = {GitHub repository},
+  url = {https://github.com/muon-spectroscopy-computational-project/pymuon-suite/},
+}</citation>
+    </citations>
```
elichad, line 81: We should cite the UEP paper here as well.

pm_uep_plot/pm_uep_plot.xml
```diff
@@ -0,0 +1,82 @@
...
+    </inputs>
+    <outputs>
+        <data label="plot from $charge_density.name" name="out_plot" format="zip" from_work_dir="out.zip"/>
+        <data label="file tree for testing" name="file_tree" format="txt" from_work_dir="tree.txt" hidden="true" />
+    </outputs>
+    <tests>
+        <test>
+            <param name="charge_density" value="Si.den_fmt" ftype="txt"/>
+            <param name="castep_log" value="Si.castep" ftype="txt" />
+            <param name="line_plots" value="[[1,1,1],[7,6,1],[1,2,3]]" />
```
elichad, line 38: This input doesn't seem to match the accepted formats in the help text - is this an array of inputs matching the third row of the table, or something else?

pm_uep_plot/pm_uep_plot.xml
```diff
@@ -0,0 +1,82 @@
...
+        touch "\${x%%.*}.yaml" &&
+        printf "chden_path: ./ \n" >>"\${x%%.*}.yaml" &&
+        ([[ ! -z "$chden_seed" ]] && printf "chden_seed: $chden_seed \n">>"\${x%%.*}.yaml" || ( >&2 echo "chden_seed empty")) &&
+        ([[ ! -z "$gw_factor" ]] && printf "gw_factor: $gw_factor \n">>"\${x%%.*}.yaml" || ( >&2 echo "gw_factor empty")) &&
+        ([[ ! -z "$line_plots" ]] && ( printf "line_plots: $line_plots \n" | sed "s/__ob__/[/g" | sed "s/__cb__/]/g" )>>"\${x%%.*}.yaml" || ( >&2 echo "line_plots empty" && exit 2)) &&
+        ([[ ! -z "$plane_plots" ]] && ( printf "plane_plots: $plane_plots \n" | sed "s/__ob__/[/g" | sed "s/__cb__/]/g" )>>"\${x%%.*}.yaml" || ( >&2 echo "plane_plots empty" && exit 2)) &&
+        cat "\${x%%.*}.yaml" &&
+        pm-uep-plot "\${x%%.*}.yaml" &&
+        ls > tree.txt &&
+        zip out.zip * -x "tree.txt"
```
elichad, line 20: Integration test: this zip is picking up a `conda_activate.log` file when I run it. Probably better to explicitly include the things you want to zip (I don't know what chaos might be caused on other instances by this)

pm_uep_plot/pm_uep_plot.xml
```diff
@@ -0,0 +1,82 @@
...
+    ]]></command>
+    <inputs>
+        <param type="data" name="charge_density" label="Charge density file (.den_fmt)" format="txt" help="The charge density file created by CASTEP during your initial DFT simulation for the structure."/>
+        <param type="data" name="castep_log" label="CASTEP log (.castep)" format="txt" help="The CASTEP log for your initial DFT simulation for the structure."/>
+        <param type="text" name="line_plots" label="Line plot coordinates" value="[]" help="Specifications for paths. Check help section for possible formats."/>
+        <param type="float" name="gw_factor" value="5.0" label="Gaussian witdh factor" help="Gaussian width factor for UEP calculation. Higher values will make the potential of atomic nuclei closer to the point-like approximation but may introduce artifacts."/>
+        <param type="text" name="chden_seed" label="charge density seed" value="" help="Seedname for the charge density calculation"/>
```
elichad, line 27: Integration test: when I set chden_seed to something other than the seed for the den_fmt and castep files, the job fails. If I set it to match the den_fmt and castep files, it's superfluous anyway since it's already calculated in the code. I don't think this parameter is necessary to include.

pm_uep_plot/pm_uep_plot.xml
```diff
@@ -0,0 +1,82 @@
...
+        <param type="text" name="plane_plots" value="[]" label="Plane plot coordinates" help="Specifications for planes. Check help section for possible formats."/>
+    </inputs>
+    <outputs>
+        <data label="plot from $charge_density.name" name="out_plot" format="zip" from_work_dir="out.zip"/>
```
elichad, line 31: So this is an underlying problem from pymuonsuite, but the tool doesn't actually output any plot, just .dat files (which I assume can be used to create one). I reckon we can improve on this - probably with a Python script to create plots. A tool that claims to plot should produce an image!

pm_uep_plot/pm_uep_plot.xml
```diff
@@ -0,0 +1,82 @@
...
+        cat "\${x%%.*}.yaml" &&
+        pm-uep-plot "\${x%%.*}.yaml" &&
+        ls > tree.txt &&
+        zip out.zip * -x "tree.txt"
+    ]]></command>
+    <inputs>
+        <param type="data" name="charge_density" label="Charge density file (.den_fmt)" format="txt" help="The charge density file created by CASTEP during your initial DFT simulation for the structure."/>
+        <param type="data" name="castep_log" label="CASTEP log (.castep)" format="txt" help="The CASTEP log for your initial DFT simulation for the structure."/>
+        <param type="text" name="line_plots" label="Line plot coordinates" value="[]" help="Specifications for paths. Check help section for possible formats."/>
+        <param type="float" name="gw_factor" value="5.0" label="Gaussian witdh factor" help="Gaussian width factor for UEP calculation. Higher values will make the potential of atomic nuclei closer to the point-like approximation but may introduce artifacts."/>
```
elichad, line 26:
```suggestion
        <param type="float" name="gw_factor" value="5.0" label="Gaussian width factor" help="Gaussian width factor for UEP calculation. Higher values will make the potential of atomic nuclei closer to the point-like approximation but may introduce artifacts."/>
```

elichad (changes requested):
Nice work to document these inputs.

Some of these comments are easy fixes, but for the stuff about UI and inputs and outputs I think we need to have a conversation with Leandro and Simone.

pm_uep_plot/pm_uep_plot.xml
Jo-stfc: the command only requires the config, but the config has the chden_path parameter which links the den_fmt and castep files

pm_uep_plot/pm_uep_plot.xml
Jo-stfc: replaced with autogenerated seed

elichad:
Discussed at team meeting. We decided to shelve this tool for a couple of reasons:

1.  its purpose (guessing stopping sites by searching for potential minima) is mostly superseded by the AIRSS method (pm-muairss + pm-uep-opt), so there isn't urgent need for it
2. the YAML input for line and plane plots is not straightforward to write - especially for a new user. This makes it hard to make a good Galaxy UI for it. We think this tool would benefit a lot from more advanced visualisations - e.g. look at a structure and visually select the lines/planes you want to study.

We'll leave the branch open and come back to this tool once we've done more visualiser work.

PR #52, opened 2022-03-25
Anish/cif2cell

### LINKED ISSUE #45: Add cif2cell tool
cif2cell has an existing conda package so this should be straightforward.

anish-mudaraddi:
Resolves #45

Running into Issue where the latest version of `cif2cell` (2.0.0a3) is only available on PyPI and not Conda - unsure how to make planemo install PyPI package 

Version on conda is 1.2.10 and uses Python2. Test with `Si.cif` fails with 
`Error: cell setup: Found niether Hall nor H-M symbol and space group 227 does not have unique setting`

This test works when using cif2cell version 2.0.0a3 available on PyPI


elichad:
I didn't realize the conda version was so out of date. You could try raising an issue on the cif2cell repository asking them to put the current version on conda https://github.com/torbjornbjorkman/cif2cell 

As a workaround, you could add a line to pip install the desired version at the start of the command section. In the long run we should definitely try and get this through conda though.

anish-mudaraddi:
Made the changes

And created a new issue on cif2cell 
https://github.com/torbjornbjorkman/cif2cell/issues/29#issue-1180886267

cif2cell/cif2cell.xml
```diff
@@ -0,0 +1,46 @@
+<tool id="cif2cell" name="Convert .cif file to .cell (cif2cell)" version="0.1.0" python_template_version="3.5">
+    <command detect_errors="exit_code"><![CDATA[
+        pip install cif2cell==2.0.0a3 &&
+        ln -s $cif_structure $cif_structure.name &&
+        cif2cell -f $cif_structure.name -p castep -o out.cell &&
+        sed -i '1,8d' out.cell
```
elichad, line 6: I don't think deleting these comments at the start of the file is a good idea – they give credit to cif2cell where it's needed. Always best to keep credit with/in the output file where possible, as these files may be taken and used elsewhere.

cif2cell/cif2cell.xml
```diff
@@ -0,0 +1,46 @@
...
+            <output name="out_cell" file="diamond_out.cell" ftype="data" />
+        </test>
+    </tests>
+    <help><![CDATA[
+        usage: cif2cell -f file.cif -p castep -o file.cell
+
+        Given an input .cif structure file, convert to equivalent .cell structure file.
+    ]]></help>
+    <citations>
+        <citation type="bibtex">
```
elichad, line 30: Not sure why, but this citation isn't appearing when I load the tool. Is it formatted correctly?

image

cif2cell/cif2cell.xml
anish-mudaraddi: The issue here is cif2cell produces a timestamp as part of the comments, so I had to delete them to make the tests pass

anish-mudaraddi:
Not sure why tests fail, it seems `.cif` is not a valid `file_type`

Same tests pass when I run `planemo test cif2cell/cif2cell.xml`

I followed the steps to checkout the Muon Galaxy branch `eli/add-crystvis`, and pointed Planemo installation to it

elichad:
I updated the GitHub workflow to point to the release_22.01 branch of Galaxy - if you rebase onto/merge with main that should hopefully fix the broken tests

elichad (approved):
Happy with this, can merge when tests pass

PR #69, opened 2022-05-23
Muspinsim Input Tool

### LINKED ISSUE #63: Add tool to create muspinsim input file
Analagous to pm_yaml_config.

anish-mudaraddi:
resolves #63 

Tool can be used to create a Muspinsim input file


anish-mudaraddi:
I needed to create a dummy .dat file: `sample_fitting_data.dat` to allow the interface for muspinsim's parsing logic to work.

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+                           help="Polar angles: Define two polar angles theta and phi, defining only the direction of the z-axis (recommended for powder averages).
+                                Euler angles: Define 3 euler angles defining a new frame, convention used is ZYZ by default.
+                                Euler angles with weight: define 3 euler angles and a weight (will be normalized automatically).
+                                Euler angles helper function: eulrange(n).
+                                Zaremba-Conroy-Wolfsberg helper function: zcw(n)">
+                        <option value="2_polar">Polar angles</option>
+                        <option value="3_euler">Euler angles</option>
+                        <option value="4_euler">Euler angles with 1 weight</option>
+                        <option value="eulrange">Euler angles helper function (eulrange(n))</option>
+                        <option value="zcw">Zaremba-Conroy-Wolfsberg helper function (zcw(n))</option>
```
elichad, line 204: Leandro tells me that zcw is the most commonly used setting as it's a convenient helper function, so this should be at the top and set as default.

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+                    </param>
+                    <repeat name="fitting_variables" title="Variable to fit to the experimental data">
+                        <param name="var_name" type="text" optional="false" label="Name of the variable"/>
+                        <param name="start_val" type="text" value="0" label="Starting value"
+                               help="Optional, default is 0. Each entry allows default expressions, constants (including MHz, muon_gyr) and functions but cannot contain names of other variables"/>
+                        <param name="min_bound" type="text" value="-inf" label="minimum bound"
+                               help="Optional, default is -inf"/>
+                        <param name="max_bound" type="text" value="inf" label="maximum bound"
+                               help="Optional, default is inf"/>
+                    </repeat>
```
elichad, line 272: This is a tricky one. Where variables are used, they will be included in other parameters earlier in the form (e.g. in this [dissipation fitting example](https://github.com/muon-spectroscopy-computational-project/muspinsim/blob/main/examples/fitting/fitting.in), so it feels weird to reference them this late. On the other hand, they definitely belong in the fitting section, which belongs at the end as it's optional. I think we leave this as-is, but perhaps be clearer in help text that variables can be used anywhere the default functions can be used (assuming this is true...)

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+                    <param name="fitting_tolerance" type="float" optional='true' value="" label="Tolerance for the fitting."
+                           help="Used as the tol parameter in Scipy's scipy.optimize.minimize method. Does not accept expressions/functions/constants"
```
elichad, line 274: Does the scipy method have a default value? If so, we should use that as the default value here (otherwise I can imagine people being confused by what they need to enter, even if it's technically optional)

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+                    />
+                </when>
+                <when value=""/>
+            </conditional>
+        </section>
+    </inputs>
+    <outputs>
+        <data format="txt" name="out_file" from_work_dir="outfile.in" />
```
elichad, line 282: This would be better with a custom label based on the `out_file_prefix` parameter - otherwise the user never actually sees the value of that parameter in the Galaxy UI. Plus if you set up multiple configs you want to be able to distinguish them!

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+            <param name="spin_preset" value="mu"/>
+            <param name="interaction" value="dissipation"/>
+            <param name="dis_index" value="1"/>
+            <param name="dis_val" value="g"/>
+            <param name="experiment_preset" value="custom" />
+            <param name="field" value="1.0/muon_gyr" />
+            <param name="fitting" value="true"/>
+            <param name="var_name" value="g" />
+            <param name="min_bound" value="0.0" />
+            <output name="out_file" file="test_5.in" ftype="txt" compare="diff" />
```
elichad, line 370: I assume all these tests correspond to muspinsim examples. Are there any variables you know to be untested in this set?

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+            <param name="spin_preset" value="custom"/>
+            <param name="spin" value="H"/>
+            <param name="spin_preset" value="mu"/>
+            <param name="spin_preset" value="e"/>
+            <param name="interaction" value="hyperfine"/>
+            <param name="hfine_index" value="2" />
+            <param name="hfine_matrix" value="[580 5  10 5   580 9 10  9   580]"/>
+            <param name="interaction" value="hyperfine"/>
+            <param name="hfine_index" value="3" />
+            <param name="hfine_matrix" value="[(300/2)  3   4*10 ] [3  15*10 6-3+2] [4  5   15 ]"/>
```
elichad, line 321: We could definitely do with a test for an input along the lines of `[[1, 2, 3], [3, 4, 5], [4, 5, 6]]` (i.e. extra brackets and comma separation + a pretty standard way to input matrices)

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+
+    Default constants include:
+        - pi: ratio of a circle and its diameter
+        - e: base of the natural logarithm
+        - deg: conversion factor between radians and degrees, equivalent to 180/pi
+        - inf: infinity
+
+    Special constants include:
+        - muon_gyr: gyromagnetic ratio of muon (135.5388 MHz/T)
+        - MHz: `1/2*muon_gyr`
```
elichad, line 396: 1. This isn't rendered correctly, or at least not as I expected (italics rather than code format, see image)
2. Is it 1/(2*muon_gyr) or (1/2)*muon_gyr?

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+        - arctan(x): inverse of the tangent
+        - arctan2(y, x): inverse of the tangent taking two arguments as (sine, cosine) to resolve the quadrant
+        - exp(x): exponential with base e
+        - log(x): natural logarithm
+        - sqrt(x): square root
+
+    Special functions include:
+        - range(x, y, z): get z equally spaced values between x and y
+        - zcw(n): Zaremba-Conroy-Wolfsberg helper function
+        - eulrange(n): helper function to create regular grid of NxNxN Euler angles with appropriate weights.
```
elichad, line 414:
```suggestion
        - eulrange(n): helper function to create regular grid of n × n × n Euler angles with appropriate weights.
```

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+
+    This tool creates a structured text file with keywords and values which describe the system to model for Muspinsim.
+    See muspinsim docs for more information https://muon-spectroscopy-computational-project.github.io/muspinsim/input/.
+
+
+    Muspinsim allows expressions and special functions to be used when defining certain keywords. This tool also allows this.
+    Check the hint at the bottom of each input to see what, if any, special function or expressions can be used.
+
+
+    Default expressions include the use of the operators \+, \-, \*, \/ and \^ for exponentiation.
```
elichad, line 385: It would be good to have these symbols in code formatting if possible

elichad (changes requested):
This tool is looking really good. I also haven't managed to break it :) I haven't tried every option and have mostly used things as intended, though. Hopefully that is/can be covered by tests.

Pretty much all of my comments are focused on UI things - wording of labels, help text etc.

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+                    </when>
+                    <when value="hyperfine">
+                        <param name="hfine_index" type="integer" value="" label="Index of non-electron spin (index starts from 1)"/>
+                        <param name="hfine_e_index" type="integer" value="" optional="true" label="Index of electron spin"
+                               help="Optional, will use first defined electron if unspecified"/>
+                        <param name="hfine_matrix" area="true" type="text" value="" label="3x3 Tensor for hyperfine coupling (MHz)"
+                               help="Define 3X3 tensor for coupling between electron and non-electron spins.
+                               Allows default expressions, constants and functions (see help)">
+                            <sanitizer>
+                                <valid initial="string.printable">
```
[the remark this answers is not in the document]
anish-mudaraddi: it's just an `area` thing - it converts newlines to `__cn__` if we don't sanitize

muspinsim_input/muspinsim_input.xml
anish-mudaraddi: Default value for `tol` is `None` - `tol` seems to do different things depending on the fitting method chosen. I will read up a bit on the fitting methods and clarify in the help text

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,423 @@
...
+                        </conditional>
+                    </repeat>
+                </when>
+            </conditional>
+            <param name="average_axes" type="select" display="checkboxes" multiple="true" optional="true" value="orientation" label="Choose keywords that should have an average carried out over them"
+                help="Any keywords not included here or as x-axis are included in separate files">
+                <option selected="true" value="orientation">orientation</option>
+                <option value="polarization">polarization</option>
+            </param>
+            <param name="euler_convention" type="select" display="radio" label="Define euler angle convention (ignored if euler angles not defined)">
```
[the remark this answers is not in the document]
anish-mudaraddi: 'orientation' is a repeating section - we only want the user to specify one euler angle convention

muspinsim_input/muspinsim_input.xml
```diff
@@ -35,46 +35,49 @@
...
                         <option value="dipolar">dipolar</option>
                         <option value="quadrupolar">quadrupolar</option>
                         <option value="dissipation">dissipation</option>
                     </param>
                     <when value="zeeman">
-                        <param name="zeeman_index" type="integer" value="" label="Index of coupled spin (index starts from 1)"/>
-                        <param name="zeeman_vector" type="text" value="" label="Vector for Zeeman coupling (Tesla)"
-                            help="Define 1X3 vector for local magnetic field coupling.
+                        <param name="zeeman_index" type="integer" value="" label="Index of coupled spin" min="1"
+                               help="Index refers to the position of the spin as selected in 'spins' section "/>
```
elichad, line 68: This is a tricky one to phrase. I think we definitely need to mention the 1-indexing though. What do you think of this?
```suggestion
                               help="Index refers to the order of the spins listed in 'spins' section.  The first spin in the list has index 1, the second has index 2, and so on. "/>
```

muspinsim_input/muspinsim_input.xml
```diff
@@ -180,72 +192,87 @@
                     </repeat>
                 </when>
             </conditional>
-            <param name="average_axes" type="select" display="checkboxes" multiple="true" optional="true" value="orientation" label="Choose keywords that should have an average carried out over them"
-                help="Any keywords not included here or as x-axis are included in separate files">
+            <param name="average_axes" type="select" display="checkboxes" multiple="true" optional="true" value="orientation" label="Average axes"
+                   help="keywords that should have an average carried out over them. Each keyword ticked should have a range specified.
+                   Keywords not ticked or set as X axis with a range will be included in separate files">
```
elichad, line 197:
```suggestion
                   Keywords not ticked or set as X axis, but which have a range set, will generate separate calculations for each value in the range when muspinsim is run.">
```

muspinsim_input/muspinsim_input.xml
```diff
@@ -180,72 +192,87 @@
...
+                        />
                     </when>
                     <when value="2_polar">
-                        <param name="theta" type="text" value="" label="theta polar angle"
-                               help="Allows default expressions, constants and functions"/>
-                        <param name="phi" type="text" value="" label="phi polar angle"/>
+                        <param name="theta" type="text" value="" label="θ (theta) angle"
+                               help="All polar angle entries allow default expressions, constants and functions"/>
+                        <param name="phi" type="text" value="" label="ϕ (phi) angle"
+                               help="All polar angle entries allow default expressions, constants and functions"/>
```
elichad, line 236: One of these is the 'azimuth' and one is the 'inclination' - but apparently [there are competing conventions for which is which](https://en.wikipedia.org/wiki/Spherical_coordinate_system#Conventions)... I assume muspinsim follows the ISO definition where theta=inclination and phi=azimuth. I think these terms should be included somewhere here.

muspinsim_input/muspinsim_input.xml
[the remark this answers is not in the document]
elichad: Ah, I see. Makes sense to move it after the orientation section as you have done, thanks

muspinsim_input/muspinsim_input.xml
```diff
@@ -102,69 +108,75 @@
...
-                <param name="experiment_preset" type="select" value='custom' display="radio" label="Choose an experiment type, or custom to define own params"
-                       help="`alc` - Avoided Level Crossing,
-                             `zero field` - No external,
-                             `custom` - No preset values fields,
-                             (See muspinsim docs 'experiment' keyword for more info)
-                            ">
+                <param name="experiment_preset" type="select" value='custom' display="radio" label="Experiment type"
+                       help="Avoided Level Crossing (ALC) - polarization: longitudinal, x-axis: field, y-axis: integral
+                             Zero field - polarization: transverse, field: 0
+                             Custom - No preset">
```
elichad, line 125: Remember newlines in the hint text don't get rendered in the UI, so extra punctuation is needed. I find the dictionary style structure here quite difficult to parse and understand here as well.

muspinsim_input/muspinsim_input.xml
```diff
@@ -102,69 +108,75 @@
...
                         <option selected="true" value="time">time</option>
                         <option value="field">field</option>
                     </param>
-                    <param name="y_axis" type="select" value="asymmetry" display="radio" label="Choose which range to use as y axis for output file(s)"
-                           help="Asymmetry: Use muon's polarization, Integral: use muon's polarization integral over time,
+                    <param name="y_axis" type="select" value="asymmetry" display="radio" label="Y axis"
+                           help="Range to use as y axis for output file(s)
+                                Asymmetry: Use muon's polarization,
+                                Integral: use muon's polarization integral over time,
                                 WARNING: if integral chosen, 'time' parameter is ignored, and cannot be used as x-axis parameter ">
```
elichad, line 149: Same comment here about line endings, for the first line in particular. This formatting should be consistent with the Experiment Type hint text - notice you've used colons in two slightly different ways, and that can be confusing.

muspinsim_input/muspinsim_input.xml
```diff
@@ -102,69 +108,75 @@
...
                         <option selected="true" value="asymmetry">asymmetry</option>
                     </param>
-                    <repeat name="fields" title="Magnetic fields to simulate">
-                        <param name="field" type="text" value="0" label="Enter field to simulate, either scalar of vector"
-                               optional="true" help="Optional, (default is 0).
-                               Accepts 1X3 vector, or scalar value. Scalar value means field is assumed to be aligned with z axis.
-                               Allows default expressions, constants and functions.
-                               Including MHz, muon_gyr and special 'range()' function (see help section)"/>
+                    <repeat name="fields" title="Fields" help="Magnetic fields">
+                         <param name="field" type="text" value="0" label="Field"
```
elichad, line 169: Don't forget units!
```suggestion
                         <param name="field" type="text" value="0" label="Field (T)"
```

muspinsim_input/muspinsim_input.xml
```diff
@@ -180,72 +192,87 @@
                     </repeat>
                 </when>
             </conditional>
-            <param name="average_axes" type="select" display="checkboxes" multiple="true" optional="true" value="orientation" label="Choose keywords that should have an average carried out over them"
-                help="Any keywords not included here or as x-axis are included in separate files">
+            <param name="average_axes" type="select" display="checkboxes" multiple="true" optional="true" value="orientation" label="Average axes"
+                   help="keywords that should have an average carried out over them. Each keyword ticked should have a range specified.
```
elichad, line 196:
```suggestion
                   help="Keywords that should have an average carried out over them. Each keyword ticked should have a range specified.
```

muspinsim_input/muspinsim_input.xml
```diff
@@ -180,72 +192,87 @@
...
-            </param>
-            <repeat name="orientations" title="Orientations to use for crystallites - (define powder averages)">
+            <repeat name="orientations" title="Orientations" help="Orientations to use for crystallites - (define powder averages)">
                 <conditional name="orientation_options">
-                    <param name="orientation_preset" type="select" display="radio" label="Choose orientation type to define" optional="false"
-                           help="Polar angles: Define two polar angles theta and phi, defining only the direction of the z-axis (recommended for powder averages).
-                                Euler angles: Define 3 euler angles defining a new frame, convention used is ZYZ by default.
-                                Euler angles with weight: define 3 euler angles and a weight (will be normalized automatically).
+                    <param name="orientation_preset" type="select" display="radio" label="Orientation" optional="false"
+                           help="Polar angles: Define two polar angles θ and ϕ, defining only the direction of the z-axis (recommended for powder averages).
```
elichad, line 207: This hint text is clear and well-formatted - even though it's long, it's pretty easy to read. This is a good example to follow for formatting.

muspinsim_input/muspinsim_input.xml
```diff
@@ -180,72 +192,87 @@
...
+                        />
                     </when>
                     <when value="zcw">
-                        <param name="zcw_n" type="text" value="0" label="value for n for zcw(n)"
-                               help="Enter number (larger numbers are more computationally expensive).
-                               09Allows default expressions, constants and functions"/>
-
+                        <param name="zcw_n" type="text" value="0" label="N"
+                               help="value for n for zcw(n)
+                               Allows default expressions, constants and functions
```
elichad, line 228:
```suggestion
                               Allows default expressions, constants and functions.
```

muspinsim_input/muspinsim_input.xml
```diff
@@ -180,72 +192,87 @@
...
-                        <option value="zcw">Zaremba-Conroy-Wolfsberg helper function (zcw(n))</option>
+                        <option selected="true" value="zcw">Zaremba-Conroy-Wolfsberg helper function (zcw(n))</option>
                     </param>
                     <when value="eulrange">
-                        <param name="eul_n" type="text" value="0" label="value for n for eulrange(n)"
-                               help="Enter number (larger numbers are more computationally expensive).
-                               Allows default expressions, constants and functions"/>
+                        <param name="eul_n" type="text" value="0" label="N"
+                               help="value for n for eulrange(n),
+                               Allows default expressions, constants and functions
```
elichad, line 221:
```suggestion
                               Allows default expressions, constants and functions.
```

muspinsim_input/muspinsim_input.xml
```diff
@@ -35,46 +35,49 @@
         <inputs name="inputs" filename="inputs.json" />
     </configfiles>
     <inputs>
-        <param type="text" name="out_file_prefix" label="A prefix for all output files (default muspinsim)" optional="true" value="muspinsim" />
-        <section name="species" expanded="true" title="Species to simulate">
-            <repeat name="spins" title="Species to simulate">
+        <param type="text" name="out_file_prefix" label="Name" help="A name with which to label this configuration" optional="true" value="muspinsim" />
+        <section name="spins" expanded="true" title="Spins" help="Section for specifying a list of the spins to be used in the system ">
+            <repeat name="spins" title="Species' spins" min="1" help="This should include a muon (mu) and one or more electrons (e)">
```
elichad, line 40: You know what, I'll backtrack on my "'to simulate' is redundant" take. In the section below, the nesting of `Spin Interactions -> Interactions to simulate -> <interaction parameters>` works well, and I think we should do the same in this section.
```suggestion
            <repeat name="spins" title="Spins to simulate" min="1" help="This should include a muon (mu) and one or more electrons (e)">
```

elichad (changes requested):
Big improvement since the last review, the form's much more readable now. Still a few smaller changes to finish things off.

PS if you agree with suggestions, you can click 'Add to batch' and commit them all at once through the GitHub UI, rather than copying changes into your local files individually. (It also gives the suggester joint credit on the commit which is nice.) The comments will auto-resolve after the commit, which is fine.

muspinsim_input/.shed.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Computational Chemistry
+- Molecular Dynamics
```
elichad, line 3: Resuggesting this now I figured out how to comment on multiple lines 😅 
```suggestion
- Materials science
- Muon spectroscopy
```

muspinsim_input/build_file.py
```diff
@@ -0,0 +1,381 @@
...
+                            "fitting_variables",
+                            [
+                                parse_fitting_variables(entry)
+                                for entry in values
+                            ],
+                        ),
+                        "fitting_tolerance": lambda value: build_block(
+                            "fitting_tolerance",
+                            [str(value)],
+                        ),
```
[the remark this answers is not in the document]
elichad: Sorry for the ambiguity - I meant the dictionary would be best placed outside the `main` function altogether. Where it is, when I start reading/skimming past it I forget what the code was trying to do by the time I get to the end 😅

muspinsim_input/muspinsim_input.xml
elichad: Ok, mentioning that Scipy will use its own defaults if the user doesn't provide a value should be sufficient.

muspinsim_input/muspinsim_input.xml
```diff
@@ -382,38 +410,52 @@
...
+        - :code:`[[1, 2, 3], [4, 5, 6], [7, 8, 9]]`
+        - :code:`1 2 3 4 5 6 7 8 9`
+        - :code:`[1 2 3] [4 5 6] [7 8 9]`
+
+
+
+    **Note** - use only square brackets when defining matrices. Accepts comma-spaced and white-spaced entries
+
+
+    **Note** - tool will automatically format matrices into appropriate dimensions, if it cannot, it will produce an error
```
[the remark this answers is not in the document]
anish-mudaraddi: I guess these Notes aren't strictly necessary - since the examples already show whats accepted - I'll remove them

muspinsim_input/muspinsim_input.xml
```diff
@@ -382,38 +410,52 @@
...
 
 
+    **Note** - to use expressions in vector/matrices please encase them in :code:`( )`
```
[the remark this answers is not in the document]
anish-mudaraddi: It's a quirk of how I interpret entries within the galaxy script

You can enter `pi` without brackets - any single function or constant works without needing brackets

technically you can also enter expressions provided there's no whitespaces in between - something like `1+2` is accepted but if you enter something like `1 + 2` with whitespaces in between it is treated as 3 separate entries. 

If you encase it in '( )' you can enter expressions and the whitespaces in between characters is automatically removed

muspinsim_input/build_file.py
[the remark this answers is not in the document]
anish-mudaraddi: So move it out into another function?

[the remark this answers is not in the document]
elichad: Since the dictionary is constant I'm thinking more of just using a variable:
```
keyword_functions = {
    "spins": lambda values: build_block(...),
    ...
}
                        

def main():
    ...
    keyword_func = keyword_functions.get(keyword, None)
    if keyword_func:
        file_contents.append(keyword_func(val))
    ...

File: muspinsim_input/muspinsim_input.xml (hunk 26 above)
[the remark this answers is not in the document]
Comment (elichad) [reply muspinsim_input/muspinsim_input.xml:458]: Ah, I see. IMO better to say what you actually want in this case:
```suggestion
    **Note** - Expressions should not contain whitespace. For example, use :code:`1+2` not :code:`1 + 2`
```
I'm willing to bet that the average physicist is much less likely to use whitespace in expressions than the average software engineer! So this should be a bit more natural for them.

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,469 @@
...
+        cp ${__tool_directory__}/sample_fitting_data.dat ./fitting_data.dat &&
+        python ${__tool_directory__}/build_file.py inputs.json
+    ]]></command>
+    <configfiles>
+        <inputs name="inputs" filename="inputs.json" />
+    </configfiles>
+    <inputs>
+        <param type="text" name="out_file_prefix" label="Name" help="A name with which to label this configuration" optional="true" value="muspinsim" />
+        <section name="spins" expanded="true" title="Spins" help="Section for specifying a list of the spins to be used in the system ">
+            <repeat name="spins" title="Spins to simulate" min="1" help="This should include a muon (mu) and one or more electrons (e)">
```
elichad, line 40: Looking at how the help text is rendered (available on `dev` branch, need to merge into others), I think it'd be better to move the section help text to the repeat. Same for the interactions section.


image

```suggestion
        <section name="spins" expanded="true" title="Spins">
            <repeat name="spins" title="Spins to simulate" min="1" help="Specify the spins to be used in the system. This should include a muon (mu) and one or more electrons (e)">
```

muspinsim_input/muspinsim_input.xml
```diff
@@ -0,0 +1,469 @@
...
+                        <param name="spin" optional="false" type="text" label="Species name"/>
+                        <param name="atomic_mass" optional="true" type="integer" min="0" value="" label="Atomic mass" help="Leave blank to use default mass - whole numbers only"/>
+                    </when>
+                    <when value="mu"/>
+                    <when value="e"/>
+                </conditional>
+            </repeat>
+        </section>
+        <section name="interaction_params" expanded="true" title="Spin Interactions" help="Interaction terms available: Zeeman, hyperfine, dipolar, quadrupolar or dissipation">
+            <repeat name="interactions" title="Interactions to simulate" help="Add couplings between spins, and/or dissipation terms, see muspinsim docs for more info">
```
elichad, line 57:
```suggestion
        <section name="interaction_params" expanded="true" title="Spin Interactions" help="">
            <repeat name="interactions" title="Interactions to simulate" help="Add couplings between spins, and/or dissipation terms. Interaction terms available: Zeeman, hyperfine, dipolar, quadrupolar or dissipation. See muspinsim docs for more info">
```

muspinsim_input/muspinsim_input.xml
elichad: Thanks for accepting the suggestion. This still needs updating in all the other interaction types too - see the macro idea above

muspinsim_input/muspinsim_input.xml
```diff
@@ -448,16 +446,43 @@
         - :code:`[1 2 3] [4 5 6] [7 8 9]`
 
 
-
-    **Note** - use only square brackets when defining matrices. Accepts comma-spaced and white-spaced entries
+    Currently, the only fitting (function minimization) algorithms available are:
```
elichad, line 449:
```suggestion
    The fitting (function minimization) algorithms available are:
```
I don't think we have plans to support more right now.

muspinsim_input/muspinsim_input.xml
elichad: Ok, I actually prefer the way you had it before. I'd just put full stops between the lines instead of commas as the only change

elichad (changes requested):

elichad (approved):
Not sure what's going on with the tests, but this looks good to me now

PR #70, opened 2022-05-26
muspinsim tool: use conda package

### LINKED ISSUE #64: Update muspinsim tool to use conda package

anish-mudaraddi:
simple fix to resolve #64

### DIFF
--- a/muspinsim/muspinsim.xml
+++ b/muspinsim/muspinsim.xml
@@ -8,14 +8,14 @@
         <!-- typical fields to update are version, month, year, and doi -->
         <token name="@TOOL_CITATION@">
             @software{muspinsim,
-                author = {Sturniolo, Simone and Liborio, Leandro and Owen, Josh and {Muon Spectroscopy Computational Project}},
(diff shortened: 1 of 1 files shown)


muspinsim/muspinsim.xml
```diff
@@ -25,9 +25,9 @@
         <organization url="https://muon-spectroscopy-computational-project.github.io/index.html" name="The Muon Spectroscopy Computational Project"/>
     </creator>
     <requirements>
+        <requirement type="package" version="1.1.0">muspinsim</requirement>
```
elichad, line 28:
```suggestion
        <requirement type="package" version="@TOOL_VERSION@">muspinsim</requirement>
```
(You can update the macro tokens if you like, but it's not strictly needed right now.)

muspinsim/muspinsim.xml
```diff
@@ -12,7 +12,7 @@
                 license = {MIT},
```
elichad, line 12: You're also an author on v1.1.0 so you need to be added to the citation, but I can't make a suggestion for the author line, for some reason.

elichad (approved):

PR #73, opened 2022-06-15
Muspinsim Tool: pick up fitting report

### LINKED ISSUE #72: muspinsim fitting report should be picked up as an output
If you run muspinsim on the CL in fitting mode (e.g. the fitting example) a file `<name>_fitreport.txt` is generated with details of the fit, but this isn't picked up by the Galaxy tool.

```
Fitting process for fitting completed
Success achieved: True
Final absolute error <|f-f_targ|>: 0.004596737036516675
Number of simulations: 40
Number of iterations: 20

====================

Values found for fitting variables:

	g = 3.0368750000000104
```

This report should be picked up as an output when fitting is done (perhaps using the [filter](https://docs.galaxyproject.org/en/master/dev/schema.html#tool-outputs-data-filter) tag).

anish-mudaraddi:
fixes #72

### DIFF
--- a/muspinsim/muspinsim.xml
+++ b/muspinsim/muspinsim.xml
@@ -87,6 +87,7 @@
                 <element name="fitting_data.dat" file="fitting_data.dat" ftype="txt" compare="sim_size" delta="20"/>
             </output_collection>
             <output name="log_out" file="fitting.log" ftype="txt" compare="sim_size" delta="100"/>
+            <output name="fit_report" file="fitting_fitreport.txt" ftype="txt" compare="sim_size" delta="100"/>
         </test>
         <test>
             <param name="mu_sim" value="multi_out.in" ftype="txt"/>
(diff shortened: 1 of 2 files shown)


muspinsim/muspinsim.xml
```diff
@@ -61,6 +61,9 @@
             <discover_datasets pattern="__name__" directory="data" format="txt"/>
         </collection>
         <data label="muspinsim log for $mu_sim.name" name="log_out" format="txt" from_work_dir="log_out.log"/>
+        <data label="fit report for $mu_sim.name" name="fit_report" format="txt" from_work_dir="fitting_data_fitreport.txt">
```
elichad, line 64: When I try this tool in the browser with the muspinsim fitting example, the fit_report file comes back empty (0 lines). I don't think this line is finding the file properly. Can you replicate this?

anish-mudaraddi: my mistake - forgot to take into account the prefix of the file changes based on the keyword `name`

PR #74, opened 2022-06-21
Remove/reduce random fluctuations in tests

### LINKED ISSUE #3: Avoid random output in tests
Tests are currently designed with some leeway for the randomly variable output of pymuon-suite; we should use tests with a specific reproducible output. 

Depends on [pymuon-suite#26](https://github.com/muon-spectroscopy-computational-project/pymuon-suite/issues/26).

### LINKED ISSUE #58: Use <filter> to hide test outputs when not testing
take example from the muspinsim_plot tool

Reference: https://docs.galaxyproject.org/en/master/dev/schema.html#tool-outputs-data-filter

elichad:
- uses `random_seed` to remove randomness in the muairss workflow tools, fixes #3 
- adds a filter to hide file tree outputs when not testing, fixes #58 
- removes `sim_size` comparisons from all tools as this is now discouraged as of 22.01
    - favour `lines_diff` if files are broadly similar but have e.g. a single timestamp or variable that may change
    - favour `compare="re_match"` OR `compare="contains"` + `has_size` where files differ on too many lines e.g. log files with timestamps on every line
- updates various test files to work with refactored tests


elichad:
Linting fails because of the DFTB URL issue, seems fine otherwise

muspinsim_plot/muspinsim_plot.xml
```diff
@@ -98,7 +102,11 @@
             <param name="linestyle" value="solid"/>
             <param name="colour" value="b"/>
             <param name="linewidth" value="1.0"/>
-            <output name="out_file_pdf" ftype="pdf" file="fitting_example.pdf" compare="sim_size"/>
+            <output name="out_file_pdf" ftype="pdf">
+                <assert_contents>
+                    <has_size value="14785"/>
```
anish-mudaraddi, line 107: All the tests pass on my local machine except for this one where the file size in marginally different 
`Expected file size of 14785+-0 found 14791`

```suggestion
                    <has_size value="14785" delta="20"/>
```

pm_asephonons/test-data/test_out.txt
```diff
@@ -7,284 +7,22 @@
...
-	16	494.1467595263244
-	17	494.14675953247183
-	18	494.14675953791385
-	19	494.14675954637937
-	20	494.1467595472212
-	21	494.14675955289727
-	22	577.44415603853
-	23	577.4441560390783
-	24	577.4441560698491
+	Frequencies (cm^-1):
```
anish-mudaraddi, line 24: Would it be worth doing this? - just to catch expected number of frequencies?
```
	Frequencies (cm^-1):
	1
	2
	3
	...
	23
        24
```

pm_asephonons/test-data/test_out.txt
```diff
@@ -7,284 +7,22 @@
...
-		1	-0.212056	-0.197045	-0.202991
-		2	 0.212056	 0.197045	 0.202991
-		3	-0.212056	-0.197045	-0.202991
-		4	 0.212056	 0.197045	 0.202991
-		5	-0.212056	-0.197045	-0.202991
-		6	 0.212056	 0.197045	 0.202991
-		7	 0.212056	 0.197045	 0.202991
-		8	-0.212056	-0.197045	-0.202991
\ No newline at end of file
+		Atom x			 y			 z
```
anish-mudaraddi, line 28: Same thing here? It might make the `test_out.txt` to long though
```
Mode 1
Atom x			 y			 z
1
2
3
4
5
6
7
8
Mode 2
...
Mode 24
```

pm_uep_opt_write/pm_uep_opt_write.xml
```diff
@@ -47,32 +47,53 @@
...
+                </assert_contents>
+            </output>
         </test>
         <test>
+            <param name="testing" value="true"/>
             <param name="muonated_structures" value="muonated-default-folder.zip" ftype="zip"/>
             <param name="charge_density" value="Si.den_fmt" ftype="txt"/>
             <param name="castep_log" value="Si.castep" ftype="txt"/>
-            <output name="file_tree" file="tree-default-folder.txt" compare="sim_size" ftype="txt" delta="10"/>
+            <output name="file_tree" file="tree-default-folder.txt" sort="true">
```
anish-mudaraddi, line 75: If we are comparing the file directly here, why do we then have another assert on size?

anish-mudaraddi:
Something else to consider here is updating the `@WRAPPER_VERSION@` to 1 for each tool changed?

elichad:
re: wrapper versions, in general yes, but in this case, the galaxy0 versions haven't been put on the dev server yet (which is what I'd consider a 'release' in lieu of uploading to the tool shed), so it doesn't hugely matter

pm_uep_opt_write/pm_uep_opt_write.xml
elichad: Good catch, will remove

pm_asephonons/test-data/test_out.txt
elichad: Yep. Duplicate lines always match the first occurrence (i.e. no point including all the 12345678 for each mode), but I'll add all the non-duplicates

elichad:
Ready for re-review. pm-asephonons failures are due to an [upstream bug](https://github.com/dftbplus/dftbplus/issues/1064), they pass on my machine though

elichad:
> re: wrapper versions, in general yes, but in this case, the galaxy0 versions haven't been put on the dev server yet (which is what I'd consider a 'release' in lieu of uploading to the tool shed), so it doesn't hugely matter

I've now put the galaxy0 versions on the dev server to try and resolve some deployment issues, so I've bumped the versions in this PR accordingly.

muspinsim_plot/muspinsim_plot.xml
```diff
@@ -79,7 +79,11 @@
             <param name="linestyle" value="solid"/>
             <param name="colour" value="k"/>
             <param name="linewidth" value="1.0"/>
-            <output name="out_file_png" file="temp_example.png" ftype="png" compare="sim_size"/>
+            <output name="out_file_png" ftype="png">
+                <assert_contents>
+                    <has_size value="52486"/>
```
bgruening, line 84: Maybe add a delta here as well?

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -79,12 +80,13 @@
                 <param name="iscastep_sel" value="true"/>
                 <param name="castep_param" value="Si.param" ftype="txt"/>
             </conditional>
-            <output name="file_tree" file="tree_castep_t_with_out_folder_and_castep_param.txt" ftype="txt" compare="sim_size" delta="350">
+            <output name="file_tree" file="tree_castep_t_with_out_folder_and_castep_param.txt" ftype="txt" sort="true">
                 <assert_contents>
-                    <has_line_matching expression="^1$"/>
+                    <has_size value="1353"/>
```
bgruening, line 85: Mh I would expect has_size without a delta more unstable than checking a value.

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -65,12 +62,16 @@
...
     <outputs>
         <data label="Muonated $structure.name using $params.name" name="muonated_structures" format="zip" from_work_dir="out_zip.zip"/>
-        <data label="File tree (testing only)" name="file_tree" format="txt" from_work_dir="tree.txt" hidden="true"/>
+        <data label="File tree (testing only)" name="file_tree" format="txt" from_work_dir="tree.txt" hidden="true">
+            <filter>(testing == "true")</filter>
+        </data>
     </outputs>
     <tests>
+        <!-- CASTEP with out_folder and castep_param specified, param as txt -->
         <test>
```
bgruening, line 75: I you have filters in the outputs it would be good to have expect_num_outputs in the test, to test the filters.

bgruening (commented):
This might also help. You can make the test file a regex and then use that. See for example here: https://github.com/galaxyproject/tools-iuc/pull/4681/files

Issue #54, opened 2022-03-29
ln -s to $structure.name fails if name has spaces


elichad:
think this should be fixable by ~adding quotes around it~ - needs checking in all tools

Update: don't think quotes are a good idea. Consider replacing spaces with underscores instead.

PR #75, opened 2022-07-11
allow spaces in file name inputs

anish-mudaraddi:
changes for #54 - account for spaces when creating links between files


elichad (approved):
Looks good to me, sorry for the delay

PR #79, opened 2022-08-03
Set up tool shed deployment

elichad:
Adds CI to deploy tools to tool shed after updates to `main` branch. For now just uses the test tool shed as our categories aren't on the main shed yet (I'll request this).

Had to flip the 'owner' fields to match my Galaxy Tool Shed account (it wasn't actually doing anything before).

Also removes URL checks from the planemo linting to get the linting to pass - deployment won't happen if it fails. Not an ideal change but don't have any other ideas to fix the problem (see https://github.com/galaxyproject/planemo/issues/1261).

### DIFF
--- a/pm_symmetry/.shed.yml
+++ b/pm_symmetry/.shed.yml
@@ -7,5 +7,5 @@ long_description: |
   Given an input structure, generates a Wyckoff points symmetry report.
   Uses the pm-symmetry tool from pymuon-suite.
 name: pm_symmetry
-owner: muon-spectroscopy-computational-project
+owner: elichad
 remote_repository_url: https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/main/pm_symmetry
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -242,3 +242,41 @@ jobs:
       id: check
       with:
         mode: check
+
+  # deploy the tools to the toolsheds (first TTS for testing)
+  deploy:
+    name: Deploy
+    needs: [setup, lint, flake8, combine_outputs]
+    if: ${{ ( github.ref == 'refs/heads/main' ) && github.repository_owner == 'muon-spectroscopy-computational-project' }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ['3.7']
+    steps:
+    - uses: actions/checkout@v2
+      with:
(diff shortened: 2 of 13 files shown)


pm_symmetry/.shed.yml
```diff
@@ -7,5 +7,5 @@ long_description: |
   Given an input structure, generates a Wyckoff points symmetry report.
   Uses the pm-symmetry tool from pymuon-suite.
 name: pm_symmetry
-owner: muon-spectroscopy-computational-project
+owner: elichad
```
bgruening, line 10: is that on purpose?

elichad: yes - we don't have a tool shed account for the group right now (no group email, etc). I initially thought whatever was here would show up as the 'owner' on the tool shed, but I realised while doing this that it needs to refer to a real account

anish-mudaraddi:
looks good to me :+1:

pm_symmetry/.shed.yml
bgruening: this owner will be part of every tool ID and you can not change it later, or its really hard.

A TS account can be any email address. It can also be changed later whenever you need to change it, but this owner-name will be persistent. For IUC its `iuc` for the Proteomics people its `galaxyp` and for the cheminformatics people its `chemteam`.

elichad: Oh, I didn't realise it couldn't be changed. How does IUC adopt pre-existing tools then? Are they new repos with the same/another name, or do you do the 'really hard' bit?

I did think of that second part, but I'd already made a test repo manually that cemented my owner-name 😬 only on the test shed though

bgruening: > Oh, I didn't realise it couldn't be changed. How does IUC adopt pre-existing tools then? Are they new repos with the same/another name, or do you do the 'really hard' bit?

We ask the original author to provide IUC admin access to the repo under the old name. 

> I did think of that second part, but I'd already made a test repo manually that cemented my owner-name grimacing only on the test shed though

Mess around with the TTS as much as you like. The problem comes when you install tools, have them in workflows etc... then the tool ID is everywhere for reproducibility reasons and then you don't want to change it. So I recommend settling with a neutral name for your community and then start deploying to the main TS.

elichad: Ok, I see. Thanks

PR #85, opened 2022-09-06
Add submission script and space removal consistency #61 #84

### LINKED ISSUE #61: script_file parameter in .yaml file
This is connected to my previous issue.

To find the muon stopping site of muonium, we need to run pm-muairss with the 'calculator' parameter in the ".yaml" files as being 'dftb+' (the default) or 'castep'. If we use 'castep', we need to add in Galaxy the possibility of loading a .param file together with the corresponding ".cell" and ".yaml" file. 

We also have the option of adding the 'script_file' parameter to the ".yaml" file. This will produce copy a submission script  into each individually generated structure’s folder for use with submission systems on HPC machines (Scarf for us). The  scripts will be generated by using a template script that take information from the literal instance of the string {seedname}, which will be replaced with the name of the structure in that folder.  This which allows us to create submission scripts for batches of CASTEP structures. 

At the moment, we need to add in Galaxy the possibility of loading a "submission script" template file together with the corresponding ".cell", ".param" and ".yaml" files. Otherwise, we get an error message when generating the muon_airss_out folder.

### LINKED ISSUE #84: castep_param whitespace removal not reflected in params.yaml
https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/bd5632424b9e1ba94d7c47c91f0e7f08c5fda15d/pm_muairss_write/pm_muairss_write.xml#L37-L39

We remove spaces in the `castep_param` file name for the purposes of linking, but then use the original file name when modifying the `params.yaml` file.

This could result in (for example) `Si.param` on disk but `castep_param: S i.param` in the `params.yaml`.

The version without spaces should be used in both instances.

patrick-austin:
Added ability to optionally specify a submission script for a job scheduler (e.g. Slurm). This is only relevant when `iscastep`. The `params` file doesn't need to contain a value for `script_file`, and if one is present it is overwritten to ensure spaces are removed from file paths.

Additionally, the treatment of spaces in `castep_param` has been addressed so that the name without spaces is what's used in `params`, rather than the raw file name.

Have added tests for both of these changes.

Closes #84 
Closes #61


pm_muairss_write/test-data/configcastep_t_with_script_file.yaml
```diff
@@ -0,0 +1,15 @@
...
+calculator: castep
+geom_force_tol: 0.05
+clustering_method: hier
+clustering_hier_t: 0.2
+out_folder: Si-out
+castep_command: castep.serial
+castep_param: test.param
+mu_symbol: H:mu
+random_seed: 7357
+script_file: s u b m i t.sh
```
bgruening, line 15: this looks not good.

bgruening (commented):
I would encourage you to not do this. This makes the tool unportable and potentially even insecure. What is your use case here?

patrick-austin:
> I would encourage you to not do this. This makes the tool unportable and potentially even insecure. What is your use case here?

#54 / #75 added whitespace removal to `castep_param` filename as it was failing when trying to `ln`. I followed the same pattern when adding in this new `input`. I added `pm_muairss_write/test-data/configcastep_t_with_script_file.yaml` to test then when we remove spaces from a submitted filename, we then update the config file so that it also uses the version without spaces. This also relates to the comments you had on #86 regarding use of `sed`.

bgruening:
Arg, I'm sorry I should write more in my sentences :)

I was referring to the SLURM integration. Please do not do this if you can avoid it. This is not portable. A tool should not interact with SLURM or be allowed to start their own jobs. Happy to discuss this in person if you like.

patrick-austin:
Ah I see, sorry for misunderstanding. @elichad mentioned that the tools are not submitting anything to SLURM. As I understand this essentially we're including a formatted a shell script in our output (which is zipped along with other data into a single `output`). A user would then take this, unzip it and be able to submit to SLURM themselves. It might be that actually we should be going about this differently (perhaps using Pulsar: muon-spectroscopy-computational-project/muon-galaxy-playbooks#6)? If so, this issue/change might be outdated and not needed at all.

bgruening:
Oh, so this is not executed at all? Ok, than its at least no security concern :)

@elichad @patrick-austin happy to chat about this and Pulsar, also in respect to the EuroScienceGateway.

bgruening:
(@patrick-austin unrelated to this PR, but do you know about https://github.com/galaxyproject/galaxy-language-server, if you develop more Galaxy tools this thing will save you a lot of time. It has linting, snippets, and testing all integrated :))

elichad (approved):
Approving in the knowledge that we might just strip the script stuff out again later (whenever we get a CASTEP/SCARF integration working within Galaxy, I guess)

Issue #78, opened 2022-07-31
various small improvements to tools


bgruening:
* [ ] Quote all text and data params, e.g. like here: https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/main/muspinsim_plot/muspinsim_plot.xml#L11
* [ ] add required files https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-required-files
* [ ] add expect_num_outputs to tests, e.g. here https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/main/muspinsim_plot/muspinsim_plot.xml#L65 [docs](https://docs.galaxyproject.org/en/latest/dev/schema.html#id92)
* [ ] single quotes instead of double quotes for security reasons: https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/main/pm_dftb_opt_write/pm_dftb_opt_write.xml#L30
* [ ] license of the Galaxy tool, or the python code shipped with it can be annotated on the `<tool` level. e.g. here https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/main/pm_dftb_opt_write/pm_dftb_opt_write.xml#L59
* [ ] type DOI is preferred instead of bibtex, e.g. here https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/main/pm_dftb_opt_write/pm_dftb_opt_write.xml#L111
* [ ] a conditional with only one optional param can most likely be modeled with this one parameter as optional="true" https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/main/pm_muairss_write/pm_muairss_write.xml#L59

elichad:
Thanks @bgruening, this is really helpful :)

patrick-austin:
Sadly a lot of these links don't point to the line they used to or the file may have been moved/renamed. I think I've managed to work out what each was referring to, and to make this easier to track I've created new issues for these bullet points using GH permalinks. I'll leave this issue open, and once the new, specific issues have been addressed close this one too.

bgruening:
@patrick-austin let me know if you need a hand. 
We have also some best practices guidelines here: https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tools

PR #86, opened 2022-09-06
Support for allpos_filename #53

### LINKED ISSUE #53: Add allpos_filename option to tools
This is a parameter in pymuon-suite that allows all muon sites to be written to a single file - useful for visualising. We should include this in Galaxy tools if possible.

Might require some shenanigans with optional outputs - I'm not sure if that's actually possible. Or maybe setting it to always be true and then retrieving it.

patrick-austin:
Added a `select` parameter to `pm_muairss_write` to control the `allpos_filename` option. By default nothing is selected, in which case we output either 1 or 2 files as normal (`out_zip.zip` and `tree.txt`). If the file extension is set (to one of the four extensions mentioned elsewhere on the tool) then an allpos file will be generated with that format and there will be 2 or 3 outputs.

Added new explicit tests for each extension. Testing without the parameter set is implicitly done by all existing, unmodified tests.

Hopefully this approach is suitable - using a `select` ensures `ase.io.write` will only recieve a format it can handle, while also allowing all of the formats that we explicitly mention/support. If it turns out that practically speaking we'd always want the allpos file to be a particular format then this could be replaced with a simple boolean control. Given the original issue mentioned having optional additional outputs I assume we want this file to be distinct from the rest of the zipped content.

Closes #53


bgruening:
@patrick-austin sorry for my ignorance. I do not know why you are using the param.xml as input, but have you considered using https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles instead? This should save you from the sed commands and make the tool slightly more interactive.

bgruening:
I'm sure I miss something :) Just wanted to ask.

anish-mudaraddi:
Combining the two tools might be something to consider. We will also need to consider doing the same for the muspinsim tool

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -64,10 +70,19 @@
             </when>
             <when value="false"/>
         </conditional>
+        <param type="select" name="allpos_ext" label="All Positions File" help="Optionally generate output file containing all muon sites using the specified file type." optional="true">
+            <option value="cell">CELL</option>
+            <option value="cif">CIF</option>
+            <option value="xyz">XYZ (standard)</option>
+            <option value="extxyz">XYZ (extendend)</option>
+        </param>
```
anish-mudaraddi, line 78: I think it might be a good idea to hide this select in a conditional - like `iscastep` 

```suggestion
       <conditional name="isallpos">
            <param type="boolean" value="false" name="allpos_sel" label="Generate output file containing all muon sites?" help="Select 'Yes' if you want to generate additional output file containing all muon sites "/>
            <when value="true">
                <param type="select" name="allpos_ext" label="All positions file type" help="Select file type for file containing all muon sites">
                    <option value="cell">CELL</option>
                    <option value="cif">CIF</option>
                    <option value="xyz">XYZ (standard)</option>
                    <option value="extxyz">XYZ (extendend)</option>
                </param>
            </when>
            <when value="false"/>
        </conditional>
```

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -106,6 +121,58 @@
             </conditional>
             <output name="file_tree" file="tree_castep_t_no_outfolder_and_castep_param.txt" ftype="txt" sort="true" />
         </test>
+        <!-- CASTEP with allpos_ext xyz -->
+        <test expect_num_outputs="3">
+            <param name="structure" value="Si.cell" ftype="cell"/>
+            <param name="testing" value="true"/>
+            <param name="params" value="configcastep_t_with_out_folder_and_castep_param.yaml" ftype="yaml"/>
+            <param name="allpos_ext" value="xyz"/>
```
anish-mudaraddi, line 129: Need to set `allpos_sel = true` for all new tests 
```suggestion
            <param name="allpos_sel" value="true"/>
            <param name="allpos_ext" value="xyz"/>
```

pm_muairss_write/pm_muairss_write.xml
```diff
@@ -64,10 +70,19 @@
...
+            <option value="cif">CIF</option>
+            <option value="xyz">XYZ (standard)</option>
+            <option value="extxyz">XYZ (extendend)</option>
+        </param>
         <param type="hidden" name="testing" label="Test mode" value="false"/>
     </inputs>
     <outputs>
         <data label="Muonated $structure.name using $params.name" name="muonated_structures" format="zip" from_work_dir="out_zip.zip"/>
+        <data label="File containing the positions of all muon sites" name="allpos_file" from_work_dir="allpos" auto_format="true">
+            <filter>(allpos_ext)</filter>
```
anish-mudaraddi, line 84: If you accept the conditional - you'll need to change this - I think it's a bit more intuitive
```suggestion
            <filter>(isallpos['allpos_sel'] == "true")</filter>
```

anish-mudaraddi (changes requested):
I think it would be a good idea to hide this select choice behind a conditional - just so it's more intuitive for users

patrick-austin:
> I think it would be a good idea to hide this select choice behind a conditional - just so it's more intuitive for users

Thanks for the in depth feedback. The documentation on best practices https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#dynamic-options recommends that boolean parameters are not used as the control of a `<conditional>`, as "The user may not expect a boolean checkbox to change the content of a form."  There are also examples of tools using a `select` as part of a conditional here: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-conditional

So in order to comply with that we'd have a conditional with a yes/no `select`, followed by another `select` with supported file extensions. It's also worth bearing in mind that one of the comments from #78, now tracked in #94, was that 

> a conditional with only one optional param can most likely be modeled with this one parameter as optional="true"

So I was actually planning further move away from conditional sections and towards use of the `optional`. To me, the main reason I support using `optional` rather than `<conditional>` (aside from simply "the documentation said so") is the reduced number of clicks. In the former case I click on the selector then the value I want, versus click twice on the yes/no selector then twice again on the subsequent, non-optional extension selector.

Ultimately though everyone will have a different idea of what's intuitive, and I don't want end up making a mountain out of a molehill here - either approach works and the ultimate impact to a user is minor in the grand scheme of things. Perhaps it would be useful to get others' opinions and just go with the most popular option? In any case I'll hold off on #94 until we're confident we've decided on the approach we want so that we're consistent across our tools if nothing else.

pm_muairss_write/pm_muairss_write.xml
elichad: following @patrick-austin's comments, I'm also not sure about adding the conditional – it seems like extra clicks and more complex code for not much gain. I do agree with the underlying implication that this param could be more clearly optional since the Galaxy UI doesn't do much to communicate it. Suggestion below
```suggestion
        <param type="select" name="allpos_ext" label="Format for file containing all muon positions (optional)" help="If set, generates an extra output file containing all the muon positions using the specified file type." optional="true">
            <option value="none" selected="true">No format (do not write a file)</option>
            <option value="cell">CELL</option>
            <option value="cif">CIF</option>
            <option value="xyz">XYZ</option>
        </param>
```

Side note: due to the way the all_pos file is written from pymuonsuite using ASE, using `allpos.xyz` as the filename will result in an extended XYZ file - see https://gitlab.com/ase/ase/-/blob/master/ase/io/formats.py#L497. So having both XYZ options isn't necessary (I actually thought using `allpos.extxyz` might break, but it seems ok... yet the line that I think makes it ok [might actually be a bug](https://gitlab.com/ase/ase/-/blob/master/ase/io/formats.py#L954)? anyway, that's off topic)

elichad: @bgruening Yes, I've been meaning to check if there was an open issue on the Galaxy repo around this. It'd be good to have some visible marker to distinguish between required/optional parameters, especially as this is common in many (most?) other web forms

elichad: I haven't found anything with a cursory search, so I'll make an issue

bgruening: Thanks! I am also not aware of any issue about that.

elichad: The `allpos_filename: allpos.$allpos_ext` part is the reason I suggested `value="xyz"` – typically, all XYZ files use the `.xyz` extension, whether they're extended XYZ or not, so that's what ASE expects to receive (this is why `.extxyz` falls through the cracks slightly). The distinction between `xyz` and `extxyz` is something we've created at a Galaxy level – partly because of the difference in sniffers, partly because certain pymuonsuite commands won't work with standard XYZ.
All that said, since using `value="extxyz"` works ok, and it it's a bit more intuitive in terms of matching the output file datatype, I think either is fine. (realistically I don't see that part of the ASE code changing, but we can always switch later if it does)

For the selector - I foolishly didn't check the UI myself and assumed the default was blank or that it would be overridden by the default. You're right that this would be confusing and we should either make it non-optional or remove the extra option I suggested. I think either way is fine as long as there's clarity somewhere (e.g. help text) that no format selected = no file written.

patrick-austin: No that makes sense, I just was copying the formats from earlier in the form we list 4 of them. I suppose all of the additional detail in an `extxyz` just ends up as the "comment" if interpreted as a "standard" `xyz`, so in that sense it's always still a `.xyz` file. If `.xyz` is conventionally used I think go with that.

elichad: @patrick-austin This conversation is basically resolved, but one of the XYZ options still needs to be removed as agreed

patrick-austin: @elichad Happy to just display a "standard" `xyz` as discussed. I'm less clear on the outcome in terms of the "optional" nature of this. Do we want to:
1. Wrap in a conditional that controls whether to generate the file, then containing a (non-optional) selector for type only
2. Avoid the "optional" flag and make an explicit "do not generate" option within the extension selector
3. Use the optional flag, and wait for the upstream issues to make the default "Nothing selected" clearer (either before or after merging this into dev/main)

My preference would be for either 2 or 3

elichad (approved):

Issue #87, opened 2022-09-08
[pm_yaml_config] supercell help and type inconsistent


patrick-austin:
Currently, the help for the `param` says "This can either be a single int, a list of three integers or a 3x3 matrix of integers". However the `type ="integer"`, so attempting to input a list of integers (or anything that's not a single integer) will fail to submit.

Either:
- [ ] Remove the help text that references the list/matrix options. Downside is it would be more restrictive.
- [ ] Change the type to be "text" like `k_points_grid` below it. Downside is you could submit anything as text, I could say the supercell was "zebra" if I really wanted (this is already the case for `k_points_grid`.
- [x] Use a regex [validator](https://docs.galaxyproject.org/en/latest/dev/schema.html#validators-for-textual-inputs-text-select) to allow free text, but only if it matches our regex. This might be the best solution, but I don't know if we can construct a check that matches integers OR lists of 3 integers OR matrix of 3*3 integers.

I don't know how valuable the ability to specify an arbitrary supercell matrix is here, so not sure what's best.

bgruening:
Galaxy offers you some validations build-in.

https://docs.galaxyproject.org/en/latest/dev/schema.html#validators-for-textual-inputs-text-select

So you could check if it's only numbers and whitespaces and a new line character.

elichad:
Good catch. The `field` param in `muspinsim_config` is similar to this and uses a text field, but the validation there is contained within the muspinsim package rather than the Galaxy tool because it's more complex (allows variables and functions and such). In this case using the built in text validation option that @bgruening suggests is probably the way to go. You definitely _could_ write a regex to catch any of multiple formats, but I don't know that it's worth it

PR #88, opened 2022-09-08
Add a minimum value of zero to relevant params

patrick-austin:
For parameters that shouldn't be negative, add in a minimum value of 0. As far as I could tell this was most of them. Didn't alter `supercell` as I think that might need to be changed to `text` anyway: #87 

muon-spectroscopy-computational-project/pymuon-suite#42 describes using a range of values, but I haven't set a `max` for any of our numerical parameters yet (which is why I'm leaving this as a draft). I had a look through CASTEP/soprano etc. documentation and I couldn't see anything which obviously indicates what a sensible upper limit for any of them would be. This might require a more thourough understanding of the physical impact of the parameters (@leandro-liborio )?


pm_yaml_config/pm_yaml_config.xml
```diff
@@ -145,7 +145,7 @@
...
             <param type="text" argument="k_points_grid" name="k_points_grid" value="[1,1,1]" label="K-points grid" multiple="true" help="List of three integer k-points. Default is [1,1,1]."/>
-            <param type="integer" argument="max_scc_steps" name="max_scc_steps" value="200" optional="true" label="Max SCC steps" help="If applicable, max number of SCC steps to perform before giving up. Default is 200 for DFTB+ and 30 for CASTEP."/>
+            <param type="integer" argument="max_scc_steps" name="max_scc_steps" value="200" min="0" optional="true" label="Max SCC steps" help="If applicable, max number of SCC steps to perform before giving up. Default is 200 for DFTB+ and 30 for CASTEP."/>
```
bgruening, line 148: @patrick-austin another thing that I see here ... if you use argment and it the same as name, you can skip "name".

"argument='--foo-bar'" will be converted to "name="foo_bar"" and in the help section a `(--foo--bar)` is added.

elichad:
@patrick-austin I think let's just go ahead without maxima for now, from our conversations it seems like 'reasonable maximum' values aren't well defined for these parameters

elichad (approved):

PR #95, opened 2022-09-09
Replace bibtex citations with DOIs

### LINKED ISSUE #93: Replace bibtex with DOI where possible
type DOI is preferred instead of bibtex, e.g. here 

https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/bd5632424b9e1ba94d7c47c91f0e7f08c5fda15d/pm_dftb_opt/pm_dftb_opt.xml#L121-L140

My understanding based on the documentation is that in cases where we have a DOI, by changing the type that is all we need (we delete the rest of the current bibtex information). Galaxy then uses the DOI to get this information from the wherever the DOI links to and caches it. Presumably this is better as it means we're not duplicating all the information, and if we update the DOI then we wouldn't need to change anything in the tool `xml` (like we would if we use bibtex).

In cases where we have no DOI, I intend to leave the Bibtex as it is.

patrick-austin:
In every case where we had a DOI, I've replaced the `bibtex` citation with a `doi` type citation.

I've checked against the dev server, as far as I can tell the only place with any inconsistency is in [pm_dftb_opt.xml](https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/compare/patrick/93-bibtex-to-doi?expand=1#diff-26f80979afa2a3fc5cb7c5488e673e1dc02a3f73600bbd1cfe1882ee4dc95011) - "DFTB+" now displays as "DFTB\mathplus". If this is a significant issue, we could leave this as a `bibtex` citation.

Also worth noting that if testing using a Galaxy instance without the fix here galaxyproject/galaxy#14225 all the DOI links will be broken. As far as I know we're still running 22.1: muon-spectroscopy-computational-project/muon-galaxy#22

Closes #93

### DIFF
--- a/cif2cell/cif2cell.xml
+++ b/cif2cell/cif2cell.xml
@@ -42,21 +42,6 @@
         Uses electronic structure program 'castep'
     ]]></help>
     <citations>
-        <citation type="bibtex">
-            @article{cif2cell,
-                author = {Torbj\"orn Bj\"orkman},
-                title = {CIF2Cell: Generating geometries for electronic structure programs},
-                journal = {Computer Physics Communications},
-                volume = {182},
-                number = {5},
-                pages = {1183 - 1186},
-                year = {2011},
-                issn = {0010-4655},
-                doi={10.1016/j.cpc.2011.01.013},
-                URL={
-                    http://www.sciencedirect.com/science/article/pii/S0010465511000336
-                }
-            }
-        </citation>
+        <citation type="doi">10.1016/j.cpc.2011.01.013</citation>
     </citations>
 </tool>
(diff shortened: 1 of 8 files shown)


bgruening:
> "DFTB+" now displays as "DFTB\mathplus". If this is a significant issue, we could leave this as a bibtex citation.

I guess we should also fix that upstream. If you like you can create an issue and we will see how we fix that in Galaxy.

patrick-austin:
> I guess we should also fix that upstream. If you like you can create an issue and we will see how we fix that in Galaxy.

Upstream issue: galaxyproject/galaxy#14700

elichad (approved):
Approved with the caveat that some things are mildly misrendered due to upstream problems. Would rather change everything together as I think we'd likely forget to update individual problem citations when the upstream bug does get fixed.

PR #96, opened 2022-09-09
Add license attribute to all tools

patrick-austin:
Add the `license` attribute to all tools. This should now appear alongside citations, requirements etc. at the bottom of the page for each tool.

For `muspinsim...` tools, this was `MIT`. For `pm...` tools, it was `GPL-3.0` but this identifier is deprecated in favour of `GPL-3.0-only`:
https://spdx.org/licenses/GPL-3.0.html
https://spdx.org/licenses/GPL-3.0-only.html

For `cif2cell`, we didn't already have a license recorded in the xml so I checked https://pypi.org/project/cif2cell/

For [pm_dftb_opt](https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/compare/patrick/92-license-attribute?expand=1#diff-26f80979afa2a3fc5cb7c5488e673e1dc02a3f73600bbd1cfe1882ee4dc95011) the software/tool itself uses GPL3, but some of the parameter sets that can be used with DFTB use https://spdx.org/licenses/CC-BY-SA-4.0.html
e.g.
https://dftb.org/parameters/download/pbc/pbc-0-3-cc
which requires that we "indicate the Licensed Material is licensed under this Public License, and include the text of, or the URI or hyperlink to, this Public License."
I've therefore left the existing disclaimer in the help, but used GPL3 as the `license` attribute seeing as that applies to the tool itself. If someone with more knowledge of licensing thinks another approach would be more suitable, I'm happy to change this.

Closes #92

PR #98, opened 2022-09-13
Add required_files to relevant tools #90

### LINKED ISSUE #90: add required files
See documentation:
https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-required-files

patrick-austin:
Added the `required_files` tag to relevant tools.
Closes #90

### DIFF
--- a/pm_dftb_opt/pm_dftb_opt.xml
+++ b/pm_dftb_opt/pm_dftb_opt.xml
@@ -27,6 +27,11 @@
         <requirement type="package" version="0.2.1">pymuonsuite</requirement>
         <requirement type="package" version="3.0">zip</requirement>
     </requirements>
+    <required_files>
+        <include type="literal" path="get_out_folder.py"/>
+        <include type="literal" path="run.sh"/>
+        <include type="prefix" path="README-"/>
+    </required_files>
     <command detect_errors="exit_code"><![CDATA[
         unzip "$muonated_structures" &&
         if test -f "params.yaml"; then echo "params.yaml present"; else echo "params.yaml missing" && exit 64; fi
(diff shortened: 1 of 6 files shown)


pm_dftb_opt/pm_dftb_opt.xml
```diff
@@ -27,6 +27,11 @@
         <requirement type="package" version="0.2.1">pymuonsuite</requirement>
         <requirement type="package" version="3.0">zip</requirement>
     </requirements>
+    <required_files>
+        <include type="literal" path="get_out_folder.py"/>
+        <include type="literal" path="run.sh"/>
+        <include type="prefix" path="README-"/>
```
bgruening, line 33: Are the readme files really needed for the function of the tool?

patrick-austin: If we don't have the `README-` files present (I renamed them and then ran the tests with `planemo`), then the `command` will fail on this line:

https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/bd5632424b9e1ba94d7c47c91f0e7f08c5fda15d/pm_dftb_opt/pm_dftb_opt.xml#L39

`cp: cannot stat '/home/kez92691/.planemo/planemo_tmp_h4wynvqy/README-pbc': No such file or directory`

As to why we have this dependence, I believe it is (at least partially) due to the fact that the `3ob` and `pbc` datasets have a different license to the tool/dftb+ itself  which requires that we "indicate the Licensed Material is licensed under this Public License, and include the text of, or the URI or hyperlink to, this Public License" (this was also relevant to #96 ):

https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/bd5632424b9e1ba94d7c47c91f0e7f08c5fda15d/pm_dftb_opt/README-3ob#L8-L10

bgruening: Interesting... Never saw that. Thanks for the explanation

bgruening (approved):

Issue #99, opened 2022-09-15
Consider combining pm_yaml_config and pm_muairss_write


anish-mudaraddi:
pm_muairss_write tool is currently written to use an external file for setting config parameters. 

the tool pm_yaml_config is used to write this file. It also  sets parameters for the UEP and read stages,

If we wanted to combine these tools – we'd have to think about whether it'd be best to put all the options in pm_muairss_write or split them up and put them in their relevant tools. 

We'll need to identify parameters used for multiple stages and how to handle them

elichad:
Adding to this: parameters like `uep_chden` don't make sense within a purely Galaxy-based workflow. They only come into play if someone wants to take their config and use it outside Galaxy. This is confusing.

> the label is suboptimal, it took me a while to work out exactly what we were doing with this input and why it was "Only required for use outside Muon Galaxy".

patrick-austin:
A possible approach to this would be to maintain a config only tool (to generate re-usable config files), and expand the "actual" tools to have the option of either reading a full config file or alternatively, interactively specifying only the relevant config settings.

In order to prevent duplication of code in this case, we could attempt refactor sections of the current config file into macros. These would then all be imported by the config tool to maintain its current functionality. The "actual" tools could then import only the macros that are relevant to their operation, and only shows these interactive options when a full config file is not provided.

This would be a more complex approach then simply refactoring directly into the main tool. However the advantage would be allowing macros to be shared between tools without duplicating code (i.e. any parameters which are relevant to multiple tools only need their xml to be defined once). It also gives the user two modes of operation (re-usable config file or interactive), however it may not be that likely that a user will want to define the config file for external use.

The following seems to be the most useful in documenting how to use macros for this purpose:
https://planemo.readthedocs.io/en/latest/writing_advanced.html#macros-reusable-elements
there are also examples here:
https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/config/sample/job_conf.xml.sample_advanced

patrick-austin:
Another motivation for actioning this is it would prevent situations where we ask the user if their config file contains certain values (if `pm_muairss_write` had an interactive interface, we could directly check the variable):
https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/6a07c4caed5ef35f35198d54ebafd7d1d6d60edb/pm_muairss_write/pm_muairss_write.xml#L61

patrick-austin:
[...]

PR #102, opened 2022-09-26
Add validators for supercell and k_points_grid #87

### LINKED ISSUE #87: [pm_yaml_config] supercell help and type inconsistent

patrick-austin:
Changed the type of `supercell` to `text` so that lists and matrices can be submitted, as the help text suggests.

Added a `regex` validator to both `supercell` and `k_points_grid` which allows only the characters needed to build a matrix:
- whitespace
- digits
- [
- ]
- ,

Note that this technically isn't sufficient to validate the input: an input of `] , 000000 , [` would pass the validator despite obviously not being a integer, list or matrix. In the issue discussion, simple validation was deemed sufficient for these purposes. It is worth noting however that because all this tool (`pm_yaml_config`) does is generate a config file, the failure that input would cause wouldn't happen until running `pm_muairss_write`. This is worth considering in the context of #99 

Finally added tests that are expected to fail when bad inputs provided.

Closes #87

### DIFF
--- a/pm_yaml_config/pm_yaml_config.xml
+++ b/pm_yaml_config/pm_yaml_config.xml
@@ -143,8 +143,12 @@
                     <param type="integer" argument="clustering_kmeans_k" name="clustering_kmeans_k" value="4" optional="true" label="Number of clusters for k-means clustering"/>
                 </when>
             </conditional>
-            <param type="integer" argument="supercell" name="supercell" value="1" label="Supercell" multiple="true" help="Supercell size and shape to use. This can either be a single int, a list of three integers or a 3x3 matrix of integers. For a single number a diagonal matrix will be generated with the integer repeated on the diagonals. For a list of three numbers a diagonal matrix will be generated where the diagonal elements are set to the list. A matrix will be used directly as is. Default is a 3x3 identity matrix."/>
-            <param type="text" argument="k_points_grid" name="k_points_grid" value="[1,1,1]" label="K-points grid" multiple="true" help="List of three integer k-points. Default is [1,1,1]."/>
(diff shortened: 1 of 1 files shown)


bgruening:
With those changes, you should always bump the tool version number, usually the galaxy0 number to galaxy1.

patrick-austin:
> With those changes, you should always bump the tool version number, usually the galaxy0 number to galaxy1.

I spoke about this with @elichad , given there are a going to a lot of small changes to the tools as part of this effort on adopting the best practices for the tool xml, we were planning on merging all those changes into a `dev` branch then bumping the wrapper versions when we merge all of those changes into `main`. I do appreciate the reminder though, as I did forget about the need to update the version number until I'd made quite a few of these PRs...

pm_yaml_config/pm_yaml_config.xml
```diff
@@ -189,6 +193,24 @@
...
+                <param name="supercell" value="bad format"/>
+            </section>
+        </test>
+        <test expect_failure="true">
+            <section name="general_params">
+            </section>
+            <section name="calculator_params">
+            </section>
+            <section name="clustering_params">
+                <param name="k_points_grid" value="bad format"/>
```
bgruening, line 211: mh, I'm surprised that these tests are now passing :(
They should fail right?

bgruening: oh gosh I have not seen that, sorry :(

Great work!

bgruening (approved):

Issue #106, opened 2022-09-26
dftb_optionals is unused


patrick-austin:
When our calculator is `dftb+` or `all`, we can define `dftb_optionals`. However, this is never used in the `command`.
https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/6a07c4caed5ef35f35198d54ebafd7d1d6d60edb/pm_yaml_config/pm_yaml_config.xml#L109
https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/6a07c4caed5ef35f35198d54ebafd7d1d6d60edb/pm_yaml_config/pm_yaml_config.xml#L125

Either we should actually use the `text` provided, or this option should be removed. It's not immediately obvious to me what this does based on the description in the tool, so I'm not sure what the course of action should be (i.e. if we do want this `param`, how should it be encoded into the config yaml).

patrick-austin:
Having had a look at:
https://github.com/muon-spectroscopy-computational-project/pymuon-suite/blob/3aaab2de4b24f40f8361950117e03b5ca6a9637c/pymuonsuite/schemas.py#L143-L144
https://github.com/muon-spectroscopy-computational-project/pymuon-suite/blob/3aaab2de4b24f40f8361950117e03b5ca6a9637c/pymuonsuite/io/dftb.py#L286-L300
This is just a list of strings with its own validation being performed in the pm code. I don't see any reason for us to not be including this alongside the rest of the config options.

PR #107, opened 2022-09-26
Add argument to cif2cell and remove redundant names #103

### LINKED ISSUE #103: Use of param attribute argument
# Context

> @patrick-austin another thing that I see here ... if you use argment and it the same as name, you can skip "name".
"argument='--foo-bar'" will be converted to "name="foo_bar"" and in the help section a `(--foo--bar)` is added.
_Originally posted by @bgruening in https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/pull/88#discussion_r966267637_

See https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param

Having looked at the documentation, this is intended to be used for **command line** arguments. There is an instance of us not using this attribute when it would be relevant (this is however slightly complicated by the fact we strip out whitespace from `$cif_structure.name` before passing it as the `file` argument):
https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/6a07c4caed5ef35f35198d54ebafd7d1d6d60edb/cif2cell/cif2cell.xml#L17-L24
```
    -f FILE, --file=FILE
                        Input CIF file, unless given as first argument to the
                        program.
```

We also use it extensively when generating the pm yaml config, despite the fact that none of these are **command line** arguments:
https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/6a07c4caed5ef35f35198d54ebafd7d1d6d60edb/pm_yaml_config/pm_yaml_config.xml#L75
However, there is some value in using `argument` here given that these `param` `name`s do correspond exactly to the yaml configuration keys, and so it may be desirable to display these after the help text on the form.

Finally, it's worth noting that we do not use `argument` at all in `muspinsim_config`. Despite fulfilling a similar role to the previous tool, here there is a more complex determination of the keys that appear in the config file. For example, a value of `alc` for `experiment_preset` will actually end up with a key/value pair of `experiment` `alc`.

# Proposed Changes
- [x] Add `argument` of `file` to cif2cell (and remove current `name`)
- Either:
  - [ ] Remove all references to `argument` from `pm_yaml_config` as technically none of these are CL arguments
  - [x] Embrace use of `argument` in `pm_yaml_config` due to it highlighting the exact keys in the help text, and remove `name`s of these `param`s as these are superfluous once `argument` is provided. 

- [x] Leave `muspinsim_config` as it is

Of the options for `pm_yaml_config`, I'd prefer the second as I think the value of improved help text outweighs not using `argument` for its intended purposes, but if other people have strong opinions either way it'd be good to here those.

patrick-austin:
# cif2cell
Rename `name="cif_structure"` to `argument="--file"` (implicitly `name="file"`) to match the command line option of cif2cell

# py_yaml_config
Rename `struct_name` to `name` to be consistent with the actual key used in the yaml output.
Remove all redundant `name="abc"`
Did not change `dftb_optionals` as that isn't even written to the yaml, and should be addressed by #106 

Closes #103

### DIFF
--- a/cif2cell/cif2cell.xml
+++ b/cif2cell/cif2cell.xml
@@ -15,23 +15,23 @@
         <requirement type="package" version="2.0.0a3">cif2cell</requirement>
     </requirements>
     <command detect_errors="exit_code"><![CDATA[
-        cif_structure_name=\$(sed 's/ //g' <<< "$cif_structure.name") &&
-        ln -s $cif_structure \$cif_structure_name &&
-        cif2cell -f \$cif_structure_name -p castep -o out.cell
+        file_name=\$(sed 's/ //g' <<< "$file.name") &&
+        ln -s $file \$file_name &&
+        cif2cell -f \$file_name -p castep -o out.cell
     ]]></command>
     <inputs>
-        <param type="data" name="cif_structure" format="cif" label="Structure file to Convert (.cif)" />
+        <param type="data" argument="--file" format="cif" label="Structure file to Convert (.cif)" />
     </inputs>
     <outputs>
-        <data label="Conversion of $cif_structure.name to .cell" name="out_cell" format="cell" from_work_dir="out.cell" />
+        <data label="Conversion of $file.name to .cell" name="out_cell" format="cell" from_work_dir="out.cell" />
     </outputs>
     <tests>
         <test expect_num_outputs="1">
-            <param name="cif_structure" value="Si.cif" ftype="cif" />
+            <param name="file" value="Si.cif" ftype="cif" />
             <output name="out_cell" file="Si_out.cell" compare="diff" lines_diff="2" ftype="cell" />
         </test>
         <test expect_num_outputs="1">
-            <param name="cif_structure" value="diamond.cif" ftype="cif" />
+            <param name="file" value="diamond.cif" ftype="cif" />
             <output name="out_cell" file="diamond_out.cell" compare="diff" lines_diff="2" ftype="cell" />
         </test>
     </tests>
--- a/pm_yaml_config/pm_yaml_config.xml
+++ b/pm_yaml_config/pm_yaml_config.xml
@@ -32,7 +32,7 @@
     <command detect_errors="exit_code"><![CDATA[
         touch outputx.yaml &&
         ([[ ! -z "$general_params.poisson_r" ]] && printf "poisson_r: $general_params.poisson_r\n">>outputx.yaml || ( >&2 echo "poisson_r empty" && exit 2)) &&
-        ([[ ! -z "$general_params.struct_name" ]] && printf "name: $general_params.struct_name\n">>outputx.yaml || ( >&2 echo "structure name is empty" && exit 2)) &&
+        ([[ ! -z "$general_params.name" ]] && printf "name: $general_params.name\n">>outputx.yaml || ( >&2 echo "structure name is empty" && exit 2)) &&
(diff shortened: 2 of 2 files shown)


pm_yaml_config/pm_yaml_config.xml
```diff
@@ -72,58 +72,58 @@
...
-                    <param type="text" argument="uep_chden" name="uep_chden" optional="true" label="Path to .den_fmt and .castep files" help="Name of the .den_fmt file containing the electronic density for an Unperturbed Electrostatic Potential optimisation. The corresponding .castep file must be with the same seedname. Only required for use outside Muon Galaxy"/>
+                    <param type="float" argument="uep_gw_factor" optional="true" value="5.0" label="Gaussian width factor" help="Gaussian width factor for UEP calculation. Higher values will make the potential of atomic nuclei closer to the point-like approximation but may introduce artifacts."/>
+                    <param type="text" argument="uep_chden" optional="true" label="Path to .den_fmt and .castep files" help="Name of the .den_fmt file containing the electronic density for an Unperturbed Electrostatic Potential optimisation. The corresponding .castep file must be with the same seedname. Only required for use outside Muon Galaxy"/>
```
bgruening, line 97: The label is suboptimal here as the user never specifies a path. Not sure what the input here should be?

bgruening: Mh, if this really takes a path an does something with it - its actually quiet dangerous.

cif2cell/cif2cell.xml
```diff
@@ -15,23 +15,23 @@
...
+        ln -s $file \$file_name &&
+        cif2cell -f \$file_name -p castep -o out.cell
     ]]></command>
     <inputs>
-        <param type="data" name="cif_structure" format="cif" label="Structure file to Convert (.cif)" />
+        <param type="data" argument="--file" format="cif" label="Structure file to Convert (.cif)" />
     </inputs>
     <outputs>
-        <data label="Conversion of $cif_structure.name to .cell" name="out_cell" format="cell" from_work_dir="out.cell" />
+        <data label="Conversion of $file.name to .cell" name="out_cell" format="cell" from_work_dir="out.cell" />
```
bgruening, line 26: This file name can be changed after the fact, at any time, this label can at some point not be true. I would not change output label, if not really needed.

patrick-austin: So it would be better to have a label as something static like "Conversion of .cif file to .cell" that doesn't reference the input file's exact name?
Nearly all of our tools access a cheetah variable for the output `label`, so if this isn't the right way to do things it will need to be changed in all the tools, not just this one.

pm_yaml_config/pm_yaml_config.xml
elichad: @patrick-austin has it right - if the generated file is used within Galaxy, anything entered in this field gets replaced when the next tool is run (replaced by the path to the `data` type input). So it shouldn't be dangerous, just weird design...

Thanks for flagging the label confusion. We should definitely discuss whether it's necessary to keep this as part of the larger issue.

cif2cell/cif2cell.xml
elichad: Huh, we didn't think about that before. We were trying to create self-descriptive output labels, but yeah that loses its value if someone changes the names of inputs later.

This is worth a review across all tools - now that we've established a convention for tool naming, the default names might work better for us than before. At the very least it sounds like using the immutable dataset numbers is a good idea.

elichad (approved):

PR #110, opened 2022-09-26
Add support for dftb_optionals #106

### LINKED ISSUE #106: dftb_optionals is unused

patrick-austin:
Write the value of dftb_optionals to the yaml config when in `all` or `dftb+` modes. Like `supercell` and `k_points_grid`, as we expect a list using square brackets, we have to `sed` to repalce `__ob__` with `[` and `__cb__` with `]`.

Added test cases for `dftb+` and `all` modes to check the default value is correctly written to file.

Closes #106


elichad (approved):

elichad:
This needs another merge/rebase, just merge it yourself whenever you do that

PR #116, opened 2022-11-09
Add pm-nq tool

### LINKED ISSUE #115: Add pm-nq tool(s)
Currently there are two modes to `pm-nq`, read and write. This is similar to `pm-muairss`. As the step required between write and read is computationally expensive, it would require integration with a Pulsar node (SCARF) before both modes can be done by one tool wrapper - see https://github.com/muon-spectroscopy-computational-project/muon-galaxy-playbooks/issues/6.

For now, follow the pattern of muairss and create distinct read/write modes that output/accept zip. However, it may be simpler to make both these modes interactive based and contained with one wrapper, rather than two separate tools. This would make future operation when read/write/both are all valid easier.

patrick-austin:
Adds support for both `pm-nq` "tasks", write and read, with one tool wrapper. This is controlled with the task parameter, and most of the remaining parameters are only relevant for "write". In read mode, the only parameters required are the structure file and a zip file (the output of "write" plus additional files from running CASTEP or DFTB+ on that file structure).

In the future, once we have integration with our compute resources, the idea would be to be able to run both read and write so that the user does not have to deal with the intermediary zip file(s).

Closes #115


pm_nq/pm_nq.xml
```diff
@@ -0,0 +1,232 @@
...
+        <organization url="https://muon-spectroscopy-computational-project.github.io/index.html" name="The Muon Spectroscopy Computational Project"/>
+    </creator>
+    <requirements>
+        <requirement type="package" version="@TOOL_VERSION@">pymuonsuite</requirement>
+        <requirement type="package" version="1.0.0">euphonic</requirement>
+        <requirement type="package" version="3.0">zip</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+        structure_name=\$(sed 's/ //g' <<< '$structure.name') &&
+        #if $task.task == 'w':
```
bgruening, line 35: @patrick-austin can you maybe give me a small introduction why this name-wrangling is needed? What do you need to archive here?

pm_nq/pm_nq.xml
```diff
@@ -0,0 +1,232 @@
...
+                ln -s '$task.script_file' \$script_file_name &&
+                printf "script_file: \$script_file_name\n">>params.yaml &&
+            #end if
+            printf 'method: $task.method\n'>>params.yaml &&
+            printf 'mu_symbol: $task.mu_symbol\n'>>params.yaml &&
+            printf 'mu_index: $task.mu_index\n'>>params.yaml &&
+            printf 'grid_n: $task.grid_n\n'>>params.yaml &&
+            printf 'sigma_n: $task.sigma_n\n'>>params.yaml &&
+            (printf 'k_points_grid: $task.k_points_grid\n' | sed "s/__ob__/[/g" | sed "s/__cb__/]/g") >>params.yaml &&
+            printf 'avgprop: $task.avgprop\n'>>params.yaml &&
```
bgruening, line 63: Would be a config-file https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-configfiles-configfile here be better?

pm_nq/pm_nq.xml
patrick-austin: Before my involved in the project, there was an issue to remove whitespace from file names using `sed` so that we could `ln -s` them #54. As we've adopted this pattern of remove the spacing and symlinking elsewhere, I've continued it in this tool as well.

I believe we can't just use the `/tmp/tmpenfof_cq/files/7/3/6/dataset_7368bdbc-979b-41d3-98b2-c4b3138090fc.dat` filepath as a lot of the tools we're using require that the various input files associated with a single sample (e.g. Si) differ only in their file extension:

- Si.cell
- Si.param
- Si.castep
- Si.phonon
- ...etc.
 
And so if we provided `/tmp/tmpenfof_cq/files/7/3/6/dataset_7368bdbc-979b-41d3-98b2-c4b3138090fc.dat` as the structure (.cell) file, it would expect to also find a file named `/tmp/tmpenfof_cq/files/7/3/6/dataset_7368bdbc-979b-41d3-98b2-c4b3138090fc.castep`  - which doesn't exist. So we symlink them to have consistent filenames so that the tools will run, which in turn requires we don't have spaces in the filename.

pm_nq/test-data/tree.txt
```diff
@@ -1,3 +1,4 @@
+conda_activate.log
```
bgruening, line 1: is this addition correct?

patrick-austin: Apparently not - when I run the tests locally it was generating `conda_activate.log` (I don't know why) so I needed to add it to get the tests to pass. But the GHA tests don't generate it `conda_activate.log` so I've removed it.

pm_nq/pm_nq.xml
```diff
@@ -0,0 +1,253 @@
...
+                <option value="r">Average</option>
+            </param>
+            <when value="w">
+                <conditional name="phonon">
+                    <param argument="phonon_source_type" type="select" display="radio" label="Phonon Type" help="The format of the phonon file.">
+                        <option value="castep">CASTEP</option>
+                        <option value="dftb+">DFTB+</option>
+                    </param>
+                    <when value="castep">
+                        <param name="phonon_source_castep" type="data" format="txt" label="Phonon File" help="Source file for phonon modes."/>
```
elichad, line 121: I assume these are the phonons for the un-muonated structure? Might be good to make that explicit.

pm_nq/pm_nq.xml
```diff
@@ -0,0 +1,253 @@
...
+        </data>
+        <data label="Averaged Property" name="averaged_property" format="txt" from_work_dir="averages.dat">
+            <filter>(task['task'] == 'r')</filter>
+        </data>
+    </outputs>
+    <tests>
+        <!-- Test CASTEP without default values -->
+        <test expect_num_outputs="2">
+            <param name="testing" value="true"/>
+            <param name="structure" value="Si.cell" ftype="cell"/>
```
elichad, line 181: Noting that this isn't a muonated structure - the tool still works, but it's not really a 'model' input. Perhaps we could use a file with a muon as created by pm-muairss?
Though that would presumably involve remaking a bunch of other test data too - don't bother if you don't think it's worth the effort.

pm_nq/pm_nq.xml
```diff
@@ -0,0 +1,253 @@
...
+            <conditional name="task">
+                <param name="task" value="r"/>
+                <param name="zipped_input" value="zipped_input.zip" ftype="zip"/>
+            </conditional>
+            <output name="averaged_property" file="averages.dat" ftype="txt"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+        Given an input muonated structure (containing a muon), displaces the muon along its normal modes and performs vibrational averaging. There are three steps to using this tool:
+          1. Generate displaced structures with `-t w` flag
```
elichad, line 238:
```suggestion
          1. Generate displaced structures with `-t w` (write) flag
```

pm_nq/pm_nq.xml
```diff
@@ -0,0 +1,253 @@
...
+          2. Calculate the property for all displaced structures using CASTEP or DFTB+
+          3. Average the proprty across all displaced structures with `-t r` flag
```
elichad, line 240:
```suggestion
          3. Average the proprty across all displaced structures with `-t r` (read) flag
```

pm_nq/pm_nq.xml
```diff
@@ -0,0 +1,253 @@
...
+                <param argument="method" type="select" display="radio" label="Method" help="Method used to calculate thermal average">
+                    <option value="independent">Independent</option>
+                    <option value="montecarlo">Monte Carlo</option>
+                </param>
+                <param argument="mu_symbol" type="text" value="H:mu" label="Muon Symbol" help="Custom CASTEP species that identifies the muon."/>
+                <param argument="mu_index" type="integer" value="-1" label="Muon Index" help="If the Muon Symbol is not present in the cell, then the atom in the cell with this index will be treated as the muon instead. Not used if the Muon Symbol is present. Defaults to the last atom in the structure."/>
+                <param argument="grid_n" type="integer" value="20" min="1" label="Grid Points" help="Number of points to sample along each phonon mode or thermal line."/>
+                <param argument="sigma_n" type="float" value="3" min="0" label="Sigma Number" help="Number of sigmas to sample in the harmonic approximation."/>
+                <expand macro="k_points_grid"/>
```
elichad, line 138: These parameters feel less significant than the ones below them, so suggest moving them down the order.

Overall I find the current order a bit confusing as it doesn't really 'flow' from one theme to the next. Perhaps some `<section>`s could help to break up all the options as well. But getting 'flow' right is a bit of trial and error.

elichad (commented):
A few questions here - consider me approving as long as you're satisfied with your own answers

pm_nq/pm_nq.xml
```diff
@@ -0,0 +1,253 @@
...
+                        <param name="phonon_source_dftb" type="data" format="binary" label="Phonon File" help="Source file for phonon modes."/>
+                        <param name="gen_file" type="data" format="txt" label="Gen File" help="GEN formatted geometry file for DFTB+."/>
+                    </when>
+                </conditional>
+                <param argument="script_file" type="data" format="txt" optional="true" label="Script File" help="For convenience, this script file is copied to all output directories (optional)."/>
+                <param argument="method" type="select" display="radio" label="Method" help="Method used to calculate thermal average">
+                    <option value="independent">Independent</option>
+                    <option value="montecarlo">Monte Carlo</option>
+                </param>
+                <param argument="mu_symbol" type="text" value="H:mu" label="Muon Symbol" help="Custom CASTEP species that identifies the muon."/>
```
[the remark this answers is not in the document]
patrick-austin: It's used before considering the calculator to identify the muons in the structure file (as well as then being used again if using CASTEP):
https://github.com/muon-spectroscopy-computational-project/pymuon-suite/blob/3aaab2de4b24f40f8361950117e03b5ca6a9637c/pymuonsuite/quantum/vibrational/average.py#L87
The help text was based on what the PyMuonSuite schema has, which isn't a whole lot, but can elaborate to make this clearer.

pm_nq/pm_nq.xml
```diff
@@ -0,0 +1,253 @@
...
+                <param argument="mu_index" type="integer" value="-1" label="Muon Index" help="If the Muon Symbol is not present in the cell, then the atom in the cell with this index will be treated as the muon instead. Not used if the Muon Symbol is present. Defaults to the last atom in the structure."/>
+                <param argument="grid_n" type="integer" value="20" min="1" label="Grid Points" help="Number of points to sample along each phonon mode or thermal line."/>
+                <param argument="sigma_n" type="float" value="3" min="0" label="Sigma Number" help="Number of sigmas to sample in the harmonic approximation."/>
+                <expand macro="k_points_grid"/>
+                <param argument="avgprop" type="select" display="radio" label="Averaged Property" help="Property to be calculated.">
+                    <option value="hyperfine">Hyperfine</option>
+                    <option value="charge">Charge</option>
+                </param>
+                <conditional name="calculator">
```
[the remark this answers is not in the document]
patrick-austin: Not that's obvious to me, but they're treated independently by the schema (load the phonons in one format to decide the displacements, then write the dispalced structures in the other).

pm_nq/pm_nq.xml
```diff
@@ -0,0 +1,253 @@
...
+            </conditional>
+            <output name="file_tree" file="tree_extra_files.txt" ftype="txt" sort="true"/>
+        </test>
+        <!-- Test DFTB+ -->
+        <test expect_num_outputs="2">
+            <param name="testing" value="true"/>
+            <param name="structure" value="ethyleneMu_opt.xyz" ftype="xyz"/>
+            <conditional name="task">
+                <param name="task" value="w"/>
+                <param name="method" value="montecarlo"/>
```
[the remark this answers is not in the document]
patrick-austin: No the choice of sampling method and calculator are independent, I just changed both at the same time so I could hit both with one test case (i.e. because I'm lazy).

pm_nq/pm_nq.xml
patrick-austin: It's actually for the muonated structure (we're displacing the muon so need its phonon modes) - I'll make that explicit

pm_nq/pm_nq.xml
patrick-austin: In the case where nothing is labelled a muon, PyMuonSuite will assume the last atom in the structure is the muon. The same thing happens for the `ethyleneMu`, as the muon is actually labelled `H` but is at the bottom of the file.

Obviously the physical validity goes out of the window since it's actually displacing `Si`, but that's not what we're testing here (just that it generates the correct file structure). I'd leave it as is (especially since SCARF is down atm and so I can't generate any new data anyway).

pm_nq/pm_nq.xml
patrick-austin: For this and the above comment, the order is the same as the PyMuonSuite schema file. But agree that if a re-order  with sections improves comprehension then it's worth doing.

PR #119, opened 2023-01-27
Reorder extension removal in pm_uep_opt #117

### LINKED ISSUE #117: pm_uep_opt fails when den_fmt has custom name
The tool wrapper is replacing (but not setting) `chden_path` and `chden_seed` in the param file to match the galaxy dataset name. ~~If these aren't present (they're not, as the config generator never sets them) then it will default to the seedname of the structure which means the workflow will fail unless you happened to name the file the same thing.~~

~~In other places we have logic required to do the "replace or set" using `sed`, so copy that and add tests to cover this case.~~

EDIT: Actually the issue is that the wrapper implicitly assumes that the dataset name will end with a file extension, e.g. `seed.den_fmt` or even `seed.txt`, and will try to remove this from the end (to define a seed name `seed`), then add it back and load (`seed.den_fmt`). This fails when `.den_fmt` is not in the name, as then the file on disk has just the `seed` (removing the non-existent extension has no effect), but we try and load `seed.den_fmt`

patrick-austin:
PyMuonSuite loads the formatted densities by concatting the "seed" with the file extension `.den_fmt`:
https://github.com/muon-spectroscopy-computational-project/pymuon-suite/blob/3aaab2de4b24f40f8361950117e03b5ca6a9637c/pymuonsuite/calculate/uep/charged.py#L76
We were defining the seed without the extension (as it should be) but implicitly assuming that the name of the dataset ended with an extension by creating a link as is. In the case where there was no extension, this fails as the actual file on disk would just be called `seed` rather than the `seed.den_fmt` pymuonsuite tries to load.

Fixed by moving the extension removal before the link

Added test with an extensionless `den_fmt` file (copy of existing file).

Fixes #117


pm_uep_opt/pm_uep_opt.xml
```diff
@@ -37,14 +37,18 @@
         unzip '$muonated_structures' &&
         if test -f "params.yaml"; then echo "params.yaml present"; else echo "params.yaml missing" && exit 64; fi &&
         if ( test -f input_structure.* ) ; then echo "input structure present"; else echo "input structure missing" && exit 64; fi &&
+        ## PyMuonSuite requires will attempt to load '<chden_seed>.den_fmt' and
```
joelvdavies, line 40:
```suggestion
        ## PyMuonSuite will attempt to load '<chden_seed>.den_fmt' and
```

joelvdavies (commented):
Fix looks good to me. I couldn't get the workflow to run locally as it complains of missing packages but that's unrelated. Otherwise should the tool wrapper version number be increased?

patrick-austin:
> Otherwise should the tool wrapper version number be increased?

There's been quite a few minor changes to the tools that have been merged into dev (best practice refactors etc.), but not main. Once we've updated the latest versions of the underlying tools, we can merge dev into main and bump anything that needs it (but suspect it'll just end up being a version 0 wrapper for the latest tool version anyway).

joelvdavies (approved):

PR #121, opened 2023-02-01
Only accept xyz when pbc is false

### LINKED ISSUE #113: pm_asephonons fails when xyz and pbc=true
Description
-----------
As a result of fixing the default value for pbc to be true (see #112 ), the test that runs with an xyz file fails. Presumably this is because it doesn't have any lattice vectors and so `pbc` are meaningless.

In the inputs, we explicitly request a `.cell` file:
https://github.com/muon-spectroscopy-computational-project/muon-galaxy-tools/blob/6a07c4caed5ef35f35198d54ebafd7d1d6d60edb/pm_asephonons/pm_asephonons.xml#L51

However I can't see a similar restriction in PyMuonSuite itself. Furthermore, we are running tests with other types of file, `cif`, `extxyz` and `xyz`. Having played with the UI, it seems that while only cell files will be picked up automatically, I can manually upload and set an `xyz` file and submit.

Proposed Changes
-------------------
Either:

1. Remove the tests with other file types, and stick with the message that only cell files are allowed
2. Relax the message and allow the file extensions that we test for, and add in a validator that ensures we cannot submit xyz and pbc at the same time

I don't know if it's also worth change PyMuonSuite to handle this early, e.g. here:
https://github.com/muon-spectroscopy-computational-project/pymuon-suite/blob/3aaab2de4b24f40f8361950117e03b5ca6a9637c/pymuonsuite/quantum/__main__.py#L113-L118
Perhaps we could only set if the atoms object has lattice vectors, otherwise we can throw early?

patrick-austin:
Since standard XYZ doesn't have lattice vectors, attempting to run `pm_asephonons` with `pbc` set to `true` will error. Refactor the dataset selector so that we only accept XYZ files when `pbc` is `false`.

In the process, also re-ordered xml to match best practice guidelines, and replaced our `printf` with the built in `configfile` xml. Added validators to the two k point parameters, and a minimum value of 0 for tolerance.

Closes #113


muon_macros.xml
```diff
@@ -1,9 +1,12 @@
 <macros>
-    <xml name="dftb+">
+    <xml name="dftb_set">
         <param type="select" name="dftb_set" value="3ob-3-1" display="radio" label="DFTB parameter set" help="The parameter set to use for DFTB+. Currently supported are: 3ob-3-1 and pbc-0-3. See help section at the bottom of the page for details.">
```
joelvdavies, line 3: Not a change made here, but where is this information? I cant see anything in the help section.

joelvdavies (commented):
Changes look good to me. Just noticed a reference to missing help text. Also I tried the V3Si.cell file I had and it errors and said to check the asephonons.out file but the zip file was empty. Not sure if its a case of it missing the file altogether or just not locating in in this case.

muon_macros.xml
patrick-austin: Think this will be a side effect of refactoring into the macros - `pm_yaml_config` and `pm_dftb_opt` have an explaination of the DFTB datasets in their main help, but we don't have that here or in `pm_nq`, but they all import the same xml which references the help text. Will change in both places.

I don't have DFTB installed on my dev VM, if you think there might be an issue I can try and get it set up so I can properly test? There might be a legitimate issue that isn't being caught with the CI.

joelvdavies: I think it would be worth checking, failing with no explanation wouldn't be great for a user.

joelvdavies (approved):

PR #123, opened 2023-02-03
Create tool for combining MSS .dat files

### LINKED ISSUE #122: Create tool for summing/combining MSS .dat output files
Based on discussion with @joelvdavies and @leandro-liborio (feel free to add anything I've missed).

### Use case:
User may have performed a number of MSS calculations, and want to combine these into a single plot. The existing plotting tool takes dat files as an input, so can be used to plot once the dat files are combined. The files in question may, or may not, overlap and align on the x axis, both in terms of range and increments (ticks).

### Requirements:
 - When ranges and x ticks are aligned, output should simply be the sum (or average?) of all input files
 - When the files have different x ranges, the output should cover the entire range
 - In regions of overlap where x ticks are not alligned (e.g. `0.1, 0.3, 0.5` va `0.2, 0.4, 0.6`) the output should be interpolated onto a sensible output range
 - Output x range should have sensible defaults, and be configurable by user (e.g. specifying number of ticks in interval)

This can likely be achieved by wrapping a standalone Python script (like `muspinsim_plot` does).

patrick-austin:
Add tool that either averages or bins then averages the MuSpinSim dat files.

Closes #122


joelvdavies (approved):
Works great!

leandro-liborio:
Further to our discussion in the meeting, I think one of the problems we have with this was that it was taking a bit to start running.  I understand that we could fix this once we put this tool on the production server. 

As regard the issue of creating a .dat.txt file... I guess we have to live with this  for now.  Lets see what happens once we have more users.

Issue #100, opened 2022-09-15
Consider combining muspinsim config tool into muspinsim tool


anish-mudaraddi:
Same as #99 but for Muspinsim tool. Much more straightforward in this case,

PR #126, opened 2023-02-07
Add cluster save options and refactor config #125

### LINKED ISSUE #125: Support clustering_save parameters
Spotted by @leandro-liborio 

Currently the wrappers for `pm_yaml_config` and `pm_muairss_read` do not support the following options:
 - `clustering_save_min`
 - `clustering_save_type`
 - `clustering_save_format`

As the default for `clustering_save_min` is false, this means we will not produce the corresponding output(s).

Allow them to be specified, and change the output of `pm_muairss_read` to include this as an (optional) output.

patrick-austin:
Adds the `clustering_save_type` and `clustering_save_format` arguments to `pm_muairss_read`. Initially I was going to add them to `pm_yaml_config`, however it was then difficult to extract the (optional) outputs with `read` - adding the parameters directly to the read form made it possible to use them in filters so that we only generate outputs when they are needed. Note that this is related to the open issues regarding the refactor of config parameters: #99 and #100.

In terms of the implementation, for `structures` we are able to use a Galaxy `list` collection which contains each structure file. For `input` this was more difficult:
 - We have heterogeneous file types (e.g. `.gen` and `.hsd` for DFTB+) which are not (currently) supported in Galaxy. These can be explicitly set as `.txt` files, but I suspect this would cause trouble if trying to run further calculations as it can be particular about the naming pattern (including extension) it expects.
 - There appeared to be a bug with `list:list` collections where they won't display in the UI (or at least I couldn't get it to work): galaxyproject/galaxy#15520

Ultimately it seemed better to just zip the whole tree for `input`. We already do this in other places anyway, and can revisit in the future if we want. 

Closes #125

### DIFF
--- a/pm_muairss_read/pm_muairss_read.xml
+++ b/pm_muairss_read/pm_muairss_read.xml
@@ -28,6 +28,7 @@
     </creator>
     <requirements>
         <requirement type="package" version="@TOOL_VERSION@">pymuonsuite</requirement>
+        <requirement type="package" version="3.0">zip</requirement>
     </requirements>
     <required_files>
         <include type="literal" path="get_out_folder.py"/>
--- a/pm_muairss_read/pm_muairss_read.xml
+++ b/pm_muairss_read/pm_muairss_read.xml
@@ -23,6 +23,7 @@
     <creator>
         <person givenName="Jyothish" familyName="Thomas" identifier="https://orcid.org/0000-0003-4724-6924"/>
         <person givenName="Eli" familyName="Chadwick" url="https://github.com/elichad" identifier="https://orcid.org/0000-0002-0035-6475"/>
+        <person givenName="Patrick" familyName="Austin" url="https://github.com/patrick-austin"/>
         <organization url="https://muon-spectroscopy-computational-project.github.io/index.html" name="The Muon Spectroscopy Computational Project"/>
     </creator>
     <requirements>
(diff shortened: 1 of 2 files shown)


bgruening:
> We have heterogeneous file types (e.g. .gen and .hsd for DFTB+) which are not (currently) supported in Galaxy. These can be explicitly set as .txt files, but I suspect this would cause trouble if trying to run further calculations as it can be particular about the naming pattern (including extension) it expects.

Can we input two different collections?
Are .gen and .hsd files always go hand-in-hand? Or are they useable alone?

patrick-austin:
> > We have heterogeneous file types (e.g. .gen and .hsd for DFTB+) which are not (currently) supported in Galaxy. These can be explicitly set as .txt files, but I suspect this would cause trouble if trying to run further calculations as it can be particular about the naming pattern (including extension) it expects.
> 
> Can we input two different collections? Are .gen and .hsd files always go hand-in-hand? Or are they useable alone?

They need to be together, the opening of the `.hsd` makes an explicit reference to the filepath of the corresponding `.gen` file:
```
Geometry = GenFormat { 
    <<< "geo_end.gen" 
} 
...
```
The other option instead of DFTB+, CASTEP, would also output one structural (`.cell`) and one parameter file (`.param`) that are used together.

bgruening:
Ok, the correct way here would be to create a composite-datatype in Galaxy. Like html is one. This way you can have one Galaxy object that consists of 2 files. Then you can have a list of these composite datatypes and the user and all tools do not need to deal with 2 files always - just with one object.

patrick-austin:
Superseded by #130

PR #129, opened 2023-02-17
Improve help text for cif2cell, pm_yaml_config #127

### LINKED ISSUE #127: Improvements to help text
`cif2cell`: typos in description of CASTEP
`pm_muairss_write`: explanation of where the random seed is used

patrick-austin:
Closes #127

### DIFF
--- a/cif2cell/cif2cell.xml
+++ b/cif2cell/cif2cell.xml
@@ -4,7 +4,7 @@
         <!-- version of underlying tool (PEP 440) -->
         <token name="@TOOL_VERSION@">2.0.0a3</token>
         <!-- version of this tool wrapper (integer) -->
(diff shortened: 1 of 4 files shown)


leandro-liborio:
This is fine. 

I have used this tool a couple of times and it is quite efficient.  However, here in muongalaxy, we are only using cif2cell in its minimal capacity.  This is: cif2cell -f file.cif -p castep -o file.cell.  We just convert a .cif file into a .cell file.  You can do many more things with cif2cell.  I am wondering about two things:

1)  There are cases where the cif2cell -f file.cif -p castep -o file.cell process fails because .cif files have problems... What do we do with that? Shouldn't we add a warning message?

2) Is it worth considering, for the future perhaps, to include the full version of cif2cell into Galaxy?

PR #135, opened 2023-07-18
Bump combine to 2.2.1, use DOIs for MSS citations

patrick-austin:
Bump MSS version for combine tool to match other MuSpinSim tools


muspinsim_combine/muspinsim_combine.xml
```diff
@@ -4,26 +4,15 @@
...
-                license = {MIT},
-                title = {{muspinsim}},
-                url = {https://github.com/muon-spectroscopy-computational-project/muspinsim},
-                version = {v2.0.2},
-                month = {1},
-                year = {2023}
-                doi = {10.5281/zenodo.7568830}
-            }
-        </token>
+        <token name="@TOOL_CITATION@">10.5281/zenodo.7733979</token>
```
bgruening, line 12: should this token not be shared with all tools instead of redefining it in all tools

patrick-austin: Thanks for the recommendation, I've refactored this into the shared macros file

PR #140, opened 2024-02-28
Changed output labels

Neonbluestoplight:
Made most output labels follow a
"unique text on ${on_string}" format in the case
of multiple outputs else removed label entirely.

A few discrepancies worth discussing:
* **muspinsim_config.xml**: label removal shows up as "MuSpinSim Configure" instead of "${tool.name} on ${on_string}", so have left it as is for now.
* **pm_dftb_opt.xml**: The "File tree (testing only)" label was left as is since I could not see it in the output.
* **pm_muairss_read.xml**: The collection label says "minimum energy structures". Changed it but doubtful
* **pm_muairss_write.xml**: labels need reviewing

### DIFF
--- a/cif2cell/cif2cell.xml
+++ b/cif2cell/cif2cell.xml
@@ -28,7 +28,7 @@
         <param argument="--cartesian" type="boolean" truevalue="--cartesian" falsevalue="" label="Do you want to generate any output in cartesian coordinates?" help="Use a cif file to convert into a structural (cell) file." />
     </inputs>
     <outputs>
-        <data label="Conversion of $file.name to .cell" name="out_cell" format="cell" from_work_dir="out.cell" />
+        <data name="out_cell" format="cell" from_work_dir="out.cell" />
     </outputs>
     <tests>
         <test expect_num_outputs="1">
--- a/muspinsim/muspinsim.xml
+++ b/muspinsim/muspinsim.xml
@@ -51,8 +51,8 @@
         <collection name="muspinsim_results" type="list" label="Muspinsim output collection">
             <discover_datasets pattern="__name__" directory="data" format="txt"/>
         </collection>
-        <data label="muspinsim log for $mu_sim.name" name="log_out" format="txt" from_work_dir="log_out.log"/>
-        <data label="fit report for $mu_sim.name" name="fit_report" format="txt" from_work_dir="fit_report.txt">
+        <data label="muspinsim log on ${on_string}" name="log_out" format="txt" from_work_dir="log_out.log"/>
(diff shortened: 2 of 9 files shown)


patrick-austin (approved):
> muspinsim_config.xml: label removal shows up as "MuSpinSim Configure" instead of "${tool.name} on ${on_string}", so have left it as is for now.

This will be because the tool doesn't actually take any `type="data"` parameters, so `${on_string}` doesn't mean anything in this context. Use of `$out_file_prefix` is probably OK since it can't be changed after execution (Dataset labels can). But regardless, I think it would probably be best to follow through on #100 - having a "tool" which just generates a config file and then a separate tool which takes that config file as a data input does not seem right to me. That's a bigger piece of work though - in the context of this PR leaving as it is is probably best.

> pm_dftb_opt.xml: The "File tree (testing only)" label was left as is since I could not see it in the output.

That's fine, the file tree is only used for tests so shouldn't have any impact on user one way or the other.

> pm_muairss_read.xml: The collection label says "minimum energy structures". Changed it but doubtful

Did you mean to write something more here? `"Minimal energy structures on ${on_string}"` is fine in my opinion.

> pm_muairss_write.xml: labels need reviewing

You can leave this one as is. Originally it and `pm_uep_opt` were separate, but to streamline the workflow `pm_muairss_write`'s funtionality was refactored into `pm_uep_opt` - so we no longer use it, and it's moved into a section for "deprecated" tools on the website. Apologies, should've mentioned this or put a comment in the file to flag it as deprecated.

So on the whole this looks good to me - I had a look at the test failure and I think it's just an intermittent failure. It's only for the push (the pull_request actions pass) so I'm happy to approve this, as long as you're happy with my answers to the points above.

Neonbluestoplight:
> Did you mean to write something more here? `"Minimal energy structures on ${on_string}"` is fine in my opinion.

No, I didn't want to add anything else. Since it was a collection of data and not a single dataset, I wasn't sure whether I should change it or anything else, that's all.

Thank you for approving the pull request! Should I merge it now or leave it as is?

patrick-austin:
> No, I didn't want to add anything else. Since it was a collection of data and not a single dataset, I wasn't sure whether I should change it or anything else, that's all.

Sure. At the top level the collection label behaves the same as far as I know. Within the collection (i.e. for each structure), the label is handled differently, in this case `pattern="__name_and_ext__"` means it will be file name and extension, which seems to be `struct_uep_min_cluster_1` etc. in practice.

> Thank you for approving the pull request! Should I merge it now or leave it as is?

Thank you for the changes, I'll merge it now.

PR #144, opened 2024-04-18
143 mudirac as galaxy tool

subindev-d:
Created galaxy tool from mudirac, just enough to run the examples given here: https://muon-spectroscopy-computational-project.github.io/mudirac/example.html


mudirac/mudirac.xml
```diff
@@ -0,0 +1,74 @@
+<tool id="mudirac" name="MuDirac" version="@TOOL_VERSION@+galaxy@WRAPPER_VERSION@" python_template_version="3.5" profile="23.0" license="MIT">
+    <description> Muonic atom Dirac equation solver </description>
```
bgruening, line 2:
```suggestion
    <description>Muonic atom Dirac equation solver</description>
```

mudirac/mudirac.xml
```diff
@@ -0,0 +1,74 @@
+<tool id="mudirac" name="MuDirac" version="@TOOL_VERSION@+galaxy@WRAPPER_VERSION@" python_template_version="3.5" profile="23.0" license="MIT">
```
bgruening, line 1:
```suggestion
<tool id="mudirac" name="MuDirac" version="@TOOL_VERSION@+galaxy@WRAPPER_VERSION@" profile="23.0" license="MIT">
```

mudirac/mudirac.xml
```diff
@@ -0,0 +1,74 @@
...
+    </macros>
+    <creator>
+        <person givenName="Subindev" familyName="Devadasan" url="https://github.com/subindev-d" identifier="https://orcid.org/0009-0001-1723-4394"/>
+        <organization url="https://muon-spectroscopy-computational-project.github.io/index.html" name="The Muon Spectroscopy Computational Project"/>
+    </creator>
+    <requirements>
+        <requirement type="package" version="1.0.1">mudirac</requirement>
+    </requirements>
+    <command detect_errors="exit_code"><![CDATA[
+    cat $input_yaml > input.in &&
```
bgruening, line 15:
```suggestion
    cat '$input_yaml' > input.in &&
```

mudirac/mudirac.xml
```diff
@@ -0,0 +1,74 @@
...
+isotope: $isotope
+xr_lines: $xr_lines
+uehling_correction: $uehling_correction
+nuclear_model: $nuclear_model
+#if $electronic_config:
+electronic_config: $electronic_config
+#end if</configfile>
+    </configfiles>
+    <inputs>
+        <param name="element" type="text" value="H" label="Element" help="Determines the nuclear charge. Can be any symbol in the periodic table up to Z=111, Roentgenium (Rg)" optional="false"/>
```
bgruening, line 29: Would be a select with all symbols better and safer for the user?

mudirac/mudirac.xml
```diff
@@ -0,0 +1,74 @@
...
+        <param name="isotope" type="integer" value="-1" label="which isotope of the element to consider" help=" Important to determine the mass of the nucleus and its size. Default is -1,
```
bgruening, line 30: adding min/max seems to be useful here.

mudirac/mudirac.xml
```diff
@@ -0,0 +1,74 @@
...
+            <param name="electronic_config" value="Au"/>
+            <output name="output" file="Au.xr.out" ftype="txt" compare="diff"/>
+        </test>
+    </tests>
+    <help><![CDATA[
+    MuDirac is a simulation software that integrates the Dirac equation for muonic atoms to compute their X-Ray transition energies;
+    Full documentation: https://muon-spectroscopy-computational-project.github.io/mudirac/
+    ]]></help>
+    <citations>
+        <citation type="doi">https://doi.org/10.1002/xrs.3212</citation>
```
bgruening, line 72:
```suggestion
        <citation type="doi">10.1002/xrs.3212</citation>
```

bgruening (commented):
Thanks @subindev-d! I added a first small review inline. It seems that the input could be more strict.

mudirac/mudirac.xml
patrick-austin: Did this for another (currently unreleased) prototype tool:
https://github.com/MaterialsGalaxy/soprano-tools/blob/c06d72485cc03e2866b7f061766e0d28d4444e1c/soprano_plotnmr/soprano_plotnmr.xml#L7-L133

patrick-austin (changes requested):
One or two minor changes but nothing functional - ran both the gold examples locally and it worked fine which is great.

Obviously this doesn't cover all the possible keyword arguments for mudirac but I think that's fine to begin with - we can add those as we decide they're needed.

Similarly, we might decide to add more structure to (e.g.) the `xr_lines` input. We could refactor this into `<repeat/>` instead of using a `,` to delimit the text input, and have specific options in the form for the start and end state of the transition (like how we handle interactions in muspinsim). However - we might also decide it's simpler to keep things as they are (a single text field is probably quicker if you know what you're doing), so I think it's right to keep things as simple as possible for now and change things if we decide it's more user friendly in the future.

mudirac/mudirac.xml
```diff
@@ -0,0 +1,195 @@
...
+        <param name="uehling_correction" type="boolean" value="false" label="Uehling correction" help="whether to turn on the Uehling correction or not" optional="true"/>
+        <param name="electronic_config" type="text" value="" label="electronic configuration to use in order to describe the negative charge background" help="Can be a full string describing the configuration (e.g. 1s2 2s2 2p2), 
+        an element symbol to represent the default configuration of that atom when neutral (e.g. C) or a mix of the two (e.g. [He] 2s2 2p2)" optional="true"/>
+    </inputs>
+    <outputs>
+        <data name="output" format="txt" label="Output of ${tool.name}" from_work_dir="input.xr.out"/>
+        <data name="log_out" label="log file of ${tool.name}" format="txt" from_work_dir="input.log"/>
+    </outputs>
+    <tests>
+        <test>
```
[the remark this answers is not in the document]
subindev-d: there is one .err also generated, I think we should considered that too.

patrick-austin (approved):

bgruening:
Cool, very nice tool! Well done @subindev-d!

PR #146, opened 2024-05-15
add option for additonal output files, state and transition matrices

subindev-d:
mudirac hat an integer keyword `ouptut`
    output == 1 will print out only the transition energies and rates in the .xr.out file;
    output == 2  will print out also each of the states in a separate ASCII file as well as the transition matrices for each line.

---------------------------------------------------------------------------------------------------------------------------------
Added the ouput keyword to mudirac galaxy tool


bgruening:
Please bump the tool version :)

subindev-d:
> Please bump the tool version :)

it has not merged into the main branch yet, in that case do we need to chage the tool version?

bgruening:
Ah, missed that. Ok sorry for the noise!

subindev-d:
> Ah, missed that. Ok sorry for the noise!

no problemo.

mudirac/mudirac.xml
```diff
@@ -163,9 +164,21 @@ electronic_config: $electronic_config
         <param name="uehling_correction" type="boolean" value="false" label="Uehling correction" help="whether to turn on the Uehling correction or not" optional="true"/>
         <param name="electronic_config" type="text" value="" label="electronic configuration to use in order to describe the negative charge background" help="Can be a full string describing the configuration (e.g. 1s2 2s2 2p2), 
         an element symbol to represent the default configuration of that atom when neutral (e.g. C) or a mix of the two (e.g. [He] 2s2 2p2)" optional="true"/>
+        <param name="output_level" type="select" label="ouput level">
+            <option value="1" selected="true"> Only transition energies and rates</option>
+            <option value="2"> Transiion energies, states and transition matrices</option>
```
patrick-austin, line 169:
```suggestion
            <option value="2"> Transition energies, states and transition matrices</option>
```

mudirac/mudirac.xml
```diff
@@ -163,9 +164,21 @@ electronic_config: $electronic_config
...
+        </param>
```
patrick-austin, line 170: Since we only have two options here, I think we could either:
- Use `display="radio"` for radio buttons rather than a drop down select box. It's not an exact science but I think general UI/UX best practice is to use radio when there are less than ~5 options.
- Alternatively, since we only have 2 options we could make this a boolean parameter to include states and matrices in the output. Could do this by setting `truevalue="2"`, `falsevalue="1"` in the xml attributes and keeping the config formatting the same.

Personally I think boolean would be the most intuitive but if you prefer the current UI we can keep it as that, just want to make sure we've considered all the possible options.

patrick-austin (changes requested):
1 typo, and identified some possible alternatives for the layout, but they're both minor and functionally this worked well when I tried it out.

I do think we should include this in the tests as well though (either as a 3rd test case, or just include the `output=2` option in the second test case where we also set uehling etc.)

mudirac/mudirac.xml
```diff
@@ -163,9 +164,18 @@ electronic_config: $electronic_config
         <param name="uehling_correction" type="boolean" value="false" label="Uehling correction" help="whether to turn on the Uehling correction or not" optional="true"/>
         <param name="electronic_config" type="text" value="" label="electronic configuration to use in order to describe the negative charge background" help="Can be a full string describing the configuration (e.g. 1s2 2s2 2p2), 
         an element symbol to represent the default configuration of that atom when neutral (e.g. C) or a mix of the two (e.g. [He] 2s2 2p2)" optional="true"/>
+        <param name='output_level' type="boolean" value="false" falsevalue="1" truevalue="2" label="Produce states and transition matrices" help="Also produce each of the states in a separate filew as well as the transition matrices for each line"/>
```
patrick-austin, line 167:
```suggestion
        <param name='output_level' type="boolean" value="false" falsevalue="1" truevalue="2" label="Produce states and transition matrices" help="Also produce each of the states in a separate file as well as the transition matrices for each line"/>
```

patrick-austin (changes requested):
Just one typo, looks good to me other than that.

patrick-austin (approved):

PR #148, opened 2024-05-31
147 generate spectrum

subindev-d:
implemented the write_spec parameter of mudirac which generated a plot:
 added the conditional parameters: spec_step, spec_linewidth and spec_expdec

### DIFF
--- a/mudirac/mudirac.xml
+++ b/mudirac/mudirac.xml
@@ -10,7 +10,11 @@
     </creator>
     <requirements>
         <requirement type="package" version="@TOOL_VERSION@">mudirac</requirement>
+        <requirement type="package" version="3.5.1">matplotlib</requirement>
     </requirements>
+    <required_files>
+        <include type="literal" path="spectrum_plot.py"/>
+    </required_files>
     <command detect_errors="exit_code"><![CDATA[
     cat '$input_yaml' > input.in &&
     mudirac input.in
--- a/mudirac/mudirac.xml
+++ b/mudirac/mudirac.xml
@@ -14,6 +14,9 @@
     <command detect_errors="exit_code"><![CDATA[
     cat '$input_yaml' > input.in &&
     mudirac input.in
+    #if $generate_spectrum.generate_spectrum_selector=='true':
+        && mkdir plots && python '${__tool_directory__}/spectrum_plot.py' 'input.spec.dat'
+    #end if
     ]]></command>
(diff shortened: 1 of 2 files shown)


mudirac/mudirac.xml
```diff
@@ -165,6 +184,18 @@ electronic_config: $electronic_config
...
         an element symbol to represent the default configuration of that atom when neutral (e.g. C) or a mix of the two (e.g. [He] 2s2 2p2)" optional="true"/>
         <param name='output_level' type="boolean" value="false" falsevalue="1" truevalue="2" label="Produce states and transition matrices" help="Also produce each of the states in a separate file as well as the transition matrices for each line"/>
+        <conditional name="generate_spectrum">
+            <param name="generate_spectrum_selector" type="select" display="radio" value="false" label="Generate spectrum" help="Generate spectrum using the transition lines found broadened with Gaussian functions">
+                <option value="false" selected="true">False</option>
+                <option value="true">True</option>
+            </param>
+            <when value="false"/>
+            <when value="true">
+                <param name="spec_step" type="float" label="Spectrum step" help="energy step for the simulated spectrum, in eV. Default is 1E2 eV." optional="true"/>
```
patrick-austin, line 194:
```suggestion
                <param name="spec_step" type="float" value="100" min="0" label="Spectrum step" help="energy step for the simulated spectrum, in eV. Default is 1E2 eV."/>
```

See other comments about allowing negative values and using explicit defaults.

patrick-austin (changes requested):
A couple of specific points about the help text of the parameters, but in terms of the Galaxy functionality it seems to work and generate plots well.

I also made some comments about using default vs optional parameters. This is just my opinion on what I find clearer, so happy to not implement that if others feel like it's clearer as is.

There were some issues when submitting negative values. A negative linewidth doesn't really make sense, but mudirac does run with it. For example, changing linewidth from 1e5 to -1e5:
Untitled
 
Untitled

More seriously, setting a negative energy step results in a job which never finishes (normally they run in seconds but I've been waiting an hour).

I think mudirac itself should probably have validation checks to prevent this, but we should also put restrictions in place here as this will prevent the job even being submitted (failing fastest).

patrick-austin (approved):
Looks good to me, I also created muon-spectroscopy-computational-project/mudirac/issues/15 to track the underlying mudirac behaviour with negative values.

PR #150, opened 2024-06-06
including element name in the output

subindev-d:
Adding element as part of the output labels

### DIFF
--- a/mudirac/mudirac.xml
+++ b/mudirac/mudirac.xml
@@ -199,7 +199,7 @@ spec_expdec: $spec_expdec
     </inputs>
     <outputs>
         <data name="output" format="tabular" label="Output of ${tool.name} (${element})" from_work_dir="input.xr.out"/>
-        <collection name="states" format="tabular" label="States" type="list">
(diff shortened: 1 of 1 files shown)


leandro-liborio (approved):

mudirac/mudirac.xml
```diff
@@ -198,18 +198,18 @@ spec_expdec: $spec_expdec
         </conditional>
     </inputs>
     <outputs>
-        <data name="output" format="tabular" label="Output of ${tool.name}" from_work_dir="input.xr.out"/>
+        <data name="output" format="tabular" label="Output of ${tool.name} (${element})" from_work_dir="input.xr.out"/>
         <collection name="states" format="tabular" label="States" type="list">
```
patrick-austin, line 202: Is there a reason to not include the element in the label for this output? As it's present in all the others should we be consistent?

```suggestion
        <collection name="states" format="tabular" label="States (${element})" type="list">
```

patrick-austin (changes requested):

patrick-austin (approved):

PR #154, opened 2024-06-24
153 add new parameters tu mudirac in galaxy

subindev-d:
added new parameters:
- max_E_iter
- max_nodes_iter
- max_state_iter
- uehling_lowcut
- uehling_highcut
- uehling_steps

### DIFF
--- a/mudirac/mudirac.xml
+++ b/mudirac/mudirac.xml
@@ -31,7 +31,18 @@
         <configfile name="input_yaml">element: $element
 isotope: $isotope
 xr_lines: $xr_lines
-uehling_correction: $uehling_correction
+uehling_correction: $uehling_correction.uehling_correction
+#if $uehling_correction.uehling_correction=='true':
+#if $uehling_correction.uehling_steps:
+uehling_steps: $uehling_correction.uehling_steps
+#end if
+#if $uehling_correction.uehling_lowcut
+uehling_lowcut: $uehling_correction.uehling_lowcut
+#end if
+#if $uehling_correction.uehling_highcut:
+uehling_highcut: $uehling_correction.uehling_highcut
+#end if
+#end if
 nuclear_model: $nuclear_model
 output: $output_level
 write_spec: $generate_spectrum.generate_spectrum_selector
@@ -194,7 +205,18 @@ spec_expdec: $spec_expdec
             <option value="SPHERE"> finite size, uniformly charged spherical nucleus </option>
             <option value="FERMI2"> Fermi 2-term charge distribution </option>
         </param>
-        <param name="uehling_correction" type="boolean" value="false" label="Uehling correction" help="whether to turn on the Uehling correction or not" optional="true"/>
+        <conditional name="uehling_correction">
+            <param name="uehling_correction" type="select" display="radio" value="false" label="Uehling correction" help="whether to turn on the Uehling correction or not">
+                <option value="false" selected="true">False</option>
+                <option value="true">True</option>
+            </param>
+            <when value="false"/>
+            <when value="true">
+                <param name="uehling_steps" type="integer" value="" label="integration steps for the Uehling potential." help=" Higher numbers will make the Uehling energy more precise but increase computation times. Default is 100." optional="true"/>
+                <param name="uehling_lowcut" type="float" value="" label="low cutoff for Uehling potential" help="under low cutoff, the radius will be considered 0. Default is 0." optional="true"/>
+                <param name="uehling_highcut" type="float" value="" label="high cutoff for Uehling potential" help="over high cutoff, the radius will be considered r > > c . Default is INFINITY." optional="true"/>
+            </when>
+        </conditional>
         <param name="electronic_config" type="text" value="" label="electronic configuration to use in order to describe the negative charge background" help="Can be a full string describing the configuration (e.g. 1s2 2s2 2p2), 
         an element symbol to represent the default configuration of that atom when neutral (e.g. C) or a mix of the two (e.g. [He] 2s2 2p2)" optional="true"/>
         <param name="max_E_iter" type="integer" value="" label="maximum number of iterations to perform when searching for the energy of a state." help="If exceeded, convergence will fail. Increase this value for slow convergences that are however progressing. Default is 100." optional="true"/>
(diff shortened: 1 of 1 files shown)


mudirac/mudirac.xml
```diff
@@ -185,9 +205,24 @@ spec_expdec: $spec_expdec
             <option value="SPHERE"> finite size, uniformly charged spherical nucleus </option>
             <option value="FERMI2"> Fermi 2-term charge distribution </option>
         </param>
-        <param name="uehling_correction" type="boolean" value="false" label="Uehling correction" help="whether to turn on the Uehling correction or not" optional="true"/>
+        <conditional name="uehling_correction">
+            <param name="uehling_correction" type="select" display="radio" value="false" label="Uehling correction" help="whether to turn on the Uehling correction or not">
```
patrick-austin, line 209:
```suggestion
            <param name="uehling_correction" type="select" display="radio" label="Uehling correction" help="whether to turn on the Uehling correction or not">
```

I _think_ value is redundant here, as `selected="true"` on the option below is what actually sets the default: https://docs.galaxyproject.org/en/latest/dev/schema.html#tool-inputs-param-option ?

mudirac/mudirac.xml
```diff
@@ -185,9 +205,24 @@ spec_expdec: $spec_expdec
...
+                <option value="false" selected="true">False</option>
+                <option value="true">True</option>
+            </param>
+            <when value="false"/>
+            <when value="true">
+                <param name="uehling_steps" type="integer" value="" label="integration steps for the Uehling potential." help=" Higher numbers will make the Uehling energy more precise but increase computation times. Default is 100." optional="true"/>
```
patrick-austin, line 215:
```suggestion
                <param name="uehling_steps" type="integer" value="100" min="0" label="integration steps for the Uehling potential." help=" Higher numbers will make the Uehling energy more precise but increase computation times. Default is 100."/>
```

We need a minimum value here, as setting to a negative number results in a vague error:
```
terminate called after throwing an instance of 'std::bad_alloc' what(): std::bad_alloc /home/kez92691/muon-galaxy/database/jobs_directory/000/640/tool_script.sh: line 10: 4011330 Aborted (core dumped) mudirac input.in
```
This should also be properly handled by the underlying mudirac code, but the best place to prevent it in Galaxy is here (before we even submit the job).

mudirac/mudirac.xml
```diff
@@ -185,9 +205,24 @@ spec_expdec: $spec_expdec
...
+                <param name="uehling_lowcut" type="float" value="" label="low cutoff for Uehling potential" help="under low cutoff, the radius will be considered 0. Default is 0." optional="true"/>
```
patrick-austin, line 216:
```suggestion
                <param name="uehling_lowcut" type="float" value="0" min="0" label="low cutoff for Uehling potential" help="under low cutoff, the radius will be considered 0. Default is 0."/>
```
A radius should not be negative.

May be associated with the following in the error log (I changed a lot of things to be negative at once to try and break the tool):
```
2024-07-01 14-18-07.979 [Err] (main) Unknown error: State is not converged
```

mudirac/mudirac.xml
```diff
@@ -185,9 +205,24 @@ spec_expdec: $spec_expdec
...
+                <param name="uehling_highcut" type="float" value="" label="high cutoff for Uehling potential" help="over high cutoff, the radius will be considered r > > c . Default is INFINITY." optional="true"/>
```
patrick-austin, line 217:
```suggestion
                <param name="uehling_highcut" type="float" value="" min="0" label="high cutoff for Uehling potential" help="over high cutoff, the radius will be considered r > > c . Default is INFINITY." optional="true"/>
```

As above, this should not be negative (but should still allow it to be unset to get the default infinite behaviour). To be honest I kind of think we should also check that it's greater than the lowcut, as that doesn't make sense to me either, but think that's a change for mudirac rather than the wrapper.

mudirac/mudirac.xml
```diff
@@ -185,9 +205,24 @@ spec_expdec: $spec_expdec
...
+        <param name="max_nodes_iter" type="integer" value="" label="maximum number of iterations to perform when searching for a starting energy value that gives a state the expected number of nodes." help="If exceeded, convergence will fail. Should generally not need to be adjusted. Default is 100." optional="true"/>
+        <param name="max_state_iter" type="integer" value="" label=" maximum number of iterations to perform when searching for a state." help=" This loop encloses both node-based and energy-based search. Once a state is converged, the program checks again that it has the correct number of nodes. 
+        If it does not, the state is stored for future use and to provide an upper or lower limit to the energy of the searches and then the process is repeated. This number represents how much can the process be repeated before failing. Should not generally need to be adjusted. Default is 100." optional="true"/>
```
patrick-austin, line 225:
```suggestion
        <param name="max_E_iter" type="integer" value="100" min="0" label="maximum number of iterations to perform when searching for the energy of a state." help="If exceeded, convergence will fail. Increase this value for slow convergences that are however progressing. Default is 100."/>
        <param name="max_nodes_iter" type="integer" value="100" min="0" label="maximum number of iterations to perform when searching for a starting energy value that gives a state the expected number of nodes." help="If exceeded, convergence will fail. Should generally not need to be adjusted. Default is 100."/>
        <param name="max_state_iter" type="integer" value="100" min="0" label=" maximum number of iterations to perform when searching for a state." help=" This loop encloses both node-based and energy-based search. Once a state is converged, the program checks again that it has the correct number of nodes. 
        If it does not, the state is stored for future use and to provide an upper or lower limit to the energy of the searches and then the process is repeated. This number represents how much can the process be repeated before failing. Should not generally need to be adjusted. Default is 100."/>
```

Negative values for a number of iterations/steps doesn't make sense.

patrick-austin (changes requested):
Most of these do not have physically valid or logical negative values (leading to unoptimised solutions or the core being dumped and the job failing). Mudirac appears not to validate this so we should. Regarding the choice of 0 for min, I did also consider whether 1 might be better but I suppose you may want to "skip" by setting it to zero even if this could/would lead to an unoptimised solution.

In all cases we should also be altering mudirac itself, I'll update https://github.com/muon-spectroscopy-computational-project/mudirac/issues/15 as well. I'd hope that when that gets resolved, any decisions re: what lower values to accept can also be applied here.

leandro-liborio:
Hi both,

I cannot see these new parameters in MuDirac in the dev-server for Galaxy.  When will you push this there?

patrick-austin (approved):

PR #155, opened 2024-07-16
Add Mudirac

subindev-d:
- Added mudirac 
- Minor fixes for Cif2Cell and pymuonsuite tools


patrick-austin (changes requested):
The functional changes are good, and have already been reviewed when merging one by one into `dev`.
For all the other tools, where we've just modified the label, strictly speaking we have [updated the wrapper without changing the underlying version](https://galaxy-iuc-standards.readthedocs.io/en/latest/best_practices/tool_xml.html#tool-versions), so I think we should bump `@WRAPPER_VERSION@`. I'd have left suggested changes for this but can't as it's outside the vicinity of the actual change on each file, but it should be simple enough.

muspinsim/muspinsim.xml
```diff
@@ -51,8 +51,8 @@
         <collection name="muspinsim_results" type="list" label="Muspinsim output collection">
             <discover_datasets pattern="__name__" directory="data" format="txt"/>
         </collection>
-        <data label="muspinsim log for $mu_sim.name" name="log_out" format="txt" from_work_dir="log_out.log"/>
-        <data label="fit report for $mu_sim.name" name="fit_report" format="txt" from_work_dir="fit_report.txt">
+        <data label="muspinsim log on ${on_string}" name="log_out" format="txt" from_work_dir="log_out.log"/>
+        <data label="fit report on ${on_string}" name="fit_report" format="txt" from_work_dir="fit_report.txt">
```
patrick-austin, line 55: Just `muspinsim.xml` missing the wrapper bump.

patrick-austin (changes requested):

muspinsim/muspinsim.xml
subindev-d: fixed it.

patrick-austin (approved):

pm_dftb_opt/pm_dftb_opt.xml
subindev-d, line 1: deprecated

PR #157, opened 2025-08-20
updated version to 1.2.1 with new keywords

IfyyAgu:


mudirac/mudirac.xml
```diff
@@ -237,6 +249,10 @@ spec_expdec: $spec_expdec
                 <param name="spec_expdec" type="float" min="0" label="Exponential decay parameter" help="Defines the sensitivity factor E_dec, in eV. If set, multiplies the entire spectrum by exp(-E/E_dec) . If unset, no decay applied." optional="true"/>
             </when>
         </conditional>
+        <param name="radius" type="float" label="Solid sphere equivalent radius. (rms radius * sqrt(5/3))" help="By default it’s set to values from Angeli and Marinova (2013)." optional="true"/>
+        <param name="reduced_mass" type="float" label="Reduced mass of the muonic atom" help="By default it’s set to values from Angeli and Marinova (2013)." optional="true"/>
+        <param name="fermi_t" type="float" label="Skin thickness for a Fermi model nucleus" help="By default it’s 2.3 fm." optional="true"/>
```
subindev-d: fermi_c and fermi_t shoudl be enabled only if the `nuclear_model` is FERMI2

mudirac/mudirac.xml
```diff
@@ -237,6 +249,10 @@ spec_expdec: $spec_expdec
                 <param name="spec_expdec" type="float" min="0" label="Exponential decay parameter" help="Defines the sensitivity factor E_dec, in eV. If set, multiplies the entire spectrum by exp(-E/E_dec) . If unset, no decay applied." optional="true"/>
             </when>
         </conditional>
+        <param name="radius" type="float" label="Solid sphere equivalent radius. (rms radius * sqrt(5/3))" help="By default it’s set to values from Angeli and Marinova (2013)." optional="true"/>
+        <param name="reduced_mass" type="float" label="Reduced mass of the muonic atom" help="By default it’s set to values from Angeli and Marinova (2013)." optional="true"/>
```
subindev-d: you mistaken it with radius, `reduced_mass` is a boolean variable

subindev-d (changes requested):

mudirac/mudirac.xml
IfyyAgu: done

mudirac/mudirac.xml
IfyyAgu: not sure if they are optional

mudirac/mudirac.xml
```diff
@@ -200,11 +212,19 @@ spec_expdec: $spec_expdec
...
-        <param name="nuclear_model" type="select" label="Model used to describe the nucleus">
-            <option value="POINT"  selected="true"> point charge </option>
-            <option value="SPHERE"> finite size, uniformly charged spherical nucleus </option>
-            <option value="FERMI2"> Fermi 2-term charge distribution </option>
-        </param>
+        <conditional name="select_nuclear_model">
+            <param name="nuclear_model" type="select" label="Model used to describe the nucleus">
+                <option value="POINT"  selected="true"> point charge </option>
+                <option value="SPHERE"> finite size, uniformly charged spherical nucleus </option>
+                <option value="FERMI2"> Fermi 2-term charge distribution </option>
```
MilanSCD: Label can be changed to "spherically symmetric 2 parameter fermi charge distribution"

mudirac/mudirac.xml
```diff
@@ -200,11 +212,19 @@ spec_expdec: $spec_expdec
...
+            </param>
+            <when value="POINT"/>
+            <when value="SPHERE"/>
+            <when value="FERMI2">
+                <param name="fermi_t" type="float" label="Skin thickness for a Fermi model nucleus" help="By default it’s 2.3 fm."/>
```
MilanSCD: this label can be changed to "Fermi model skin thickness (fm)"

mudirac/mudirac.xml
```diff
@@ -200,11 +212,19 @@ spec_expdec: $spec_expdec
...
+                <param name="fermi_c" type="float" label="Fermi parameter c"/>
```
MilanSCD: This label can be changed to "Fermi model half density radius (fm)"

mudirac/mudirac.xml
```diff
@@ -276,6 +298,10 @@ spec_expdec: $spec_expdec
             <param name="max_E_iter" value="100"/>
             <param name="max_nodes_iter" value="100"/>
             <param name="max_state_iter" value="100"/>
+            <param name="radius" value="6.68"/>
+            <param name="reduced_mass" value="true"/>
+            <param name="fermi_t" value="2.3"/>
+            <param name="fermi_c" value="0.1"/>
```
MilanSCD: the default value for fermi_c in MuDirac is -1 this allows the fermi_c to be calculated from rms radius data tables. 
Setting a value of -1 in the input config file might be confusing for the user.
I think it should have no default value in the xml. If the param fermi_c is blank, it should not be appended to the config file.

MilanSCD (changes requested):
I think the default values for fermi_c should be removed and parameter labels could be updated as commented.

mudirac/mudirac.xml
IfyyAgu: Is this for Fermi2?

mudirac/mudirac.xml
IfyyAgu: Can you explain a bit more? I don't have a default value for fermi parameters. This is for the test section. I need to correct it because I made changes to them. They were initially independent parameters but now dependent on fermi2.

mudirac/mudirac.xml
MilanSCD: Yes so "Fermi 2-term charge distribution" could be changed. Its a minor thing though just for if we add any further models later.

mudirac/mudirac.xml
MilanSCD: Apologies I missed that it was in a test. Any positive value should be fine for testing. Typically c is between 3 and 7.

